Issue #4, opened 2026-08-13
[Fix] Complete English localization across the frontend


PierreStaub:
## Context

Lumio already provides English and German frontend dictionaries. Both currently contain the same 1,960 keys, but selecting English does not yet result in a fully English interface.

Some user-facing German strings remain hardcoded in frontend components, especially in the super-admin area. Several dates, numbers, currencies, and sorting operations are also forced to the German locale (`de-DE`), regardless of the selected language.

## Objective

Make the complete frontend usable in English while preserving the existing German interface and all current application behavior.

### Included

- Studio and authentication interfaces
- Customer galleries and other public frontend pages
- The complete super-admin interface
- Visible labels, placeholders, error messages, confirmations, tooltips, and accessibility text
- Migration of remaining hardcoded user-facing strings to the existing TypeScript dictionaries
- Locale-aware dates, numbers, currencies, and sorting
- Synchronization of the document `lang` attribute with the active locale
- American English, consistent with the existing English dictionary

### Out of scope

- Emails and other messages generated by the API
- Documentation, which already follows a separate bilingual convention
- Code comments, internal logs, and developer-facing diagnostics
- Translation of user-provided content
- Additional languages or an external localization service
- Unrelated frontend refactoring

## Functional requirements

- English must remain the default and fallback locale.
- Changing the language must update every user-facing frontend string covered by this scope.
- German must remain available and retain its existing behavior.
- Locale-sensitive values must follow the active interface language instead of a hardcoded locale.
- Destructive super-admin actions must retain their current confirmations and behavior.

## Technical requirements

- Keep the current lightweight dictionary-based i18n architecture.
- Add every new translation key to both `en.ts` and `de.ts`.
- Preserve interpolation variables and dictionary nesting.
- Replace user-facing literals with `t(...)` lookups where appropriate.
- Use the active locale for `Intl`, `toLocale*`, and locale-aware sorting.
- Keep `document.documentElement.lang` synchronized with the selected locale.
- Do not introduce data migrations, API changes, or new runtime dependencies.

## Acceptance criteria

- [ ] With English selected, no German user-facing string remains in the Studio interface.
- [ ] With English selected, no German user-facing string remains in customer galleries or public frontend pages.
- [ ] With English selected, no German user-facing string remains in the super-admin interface.
- [ ] Labels, placeholders, errors, confirmations, tooltips, and accessibility text are localized.
[...]

PR #6, opened 2026-08-14
feat(i18n): add Italian translation

### LINKED ISSUE #5: Italian Translation
Add Italian translation following contribution conventions

manuzzi:
## Summary
- Adds `apps/frontend/src/lib/i18n/it.ts` — full Italian translation of the frontend dictionary (2031/2031 keys, verified 1:1 against `en.ts`: same key names, same nesting, same section order/line count), informal "tu" register throughout.
- Registers the `it` locale in `dict.ts` and `SUPPORTED` in `i18n.tsx`, and updates the two components carrying the locale union directly (`GalleryShell.tsx`, `app/studio/settings/page.tsx`), including the human-readable "Italiano" label in both language pickers.
- Followed the "Adding a translation" section of CONTRIBUTING.md step by step.

## Notes
- `npx tsc --noEmit` passes clean.
- While reviewing rendered pages I noticed `login/page.tsx:479` has a hardcoded German string ("Passwort vergessen?") that bypasses the dictionary entirely — it shows up regardless of locale (en/de/it). Left untouched since it's out of scope for a translation-only PR; it looks like exactly the kind of thing #4 is tracking.
- Business/legal facts that are genuinely about the product's DE origin (e.g. the default 19%/7% VAT copy, "mandatory in Germany") were translated literally rather than localized to Italy, matching how `de.ts` itself already reads for its own market — this is a translation, not a re-localization of those defaults.

## Test plan
- [x] `npx tsc --noEmit` in `apps/frontend`
- [x] Ran the dev server and drove it with Playwright: `/login` and `/auth/forgot-password` with `lumio_locale=it` render the new strings correctly, no console errors tied to `it.ts`
- [ ] Would appreciate a native-speaker pass on tone/terminology from a maintainer or another contributor, especially in the super-admin and print-shop sections which I don't use day to day

Closes #5


markusthiel:
This is careful work — the consistent informal "tu", the 1:1 structural check against `en.ts`, the Playwright pass, and the fact that you spotted the hardcoded "Passwort vergessen?" in `login/page.tsx` and correctly left it alone as out of scope for a translation-only PR. I want this merged.

One thing has to happen first, and I'm sorry about the timing: `main` moved a lot while you were working. `v0.60.0` shipped the rest of #4, which added roughly 470 keys — most of them in the super-admin area, which had almost no `t()` usage before. Your branch is based on `84dc1de`, so `it.ts` is complete against the dictionary as it was, but no longer against the dictionary as it is.

Concretely, measured against current `main`:

- `en.ts` now has **2433** keys; your `it.ts` has **1960**
- **473 missing**, **0 extra** — nothing you wrote is wrong or obsolete, it is purely a gap

Where the gap sits:

| Section | Missing |
|---|---:|
| `super` | 397 |
| `studio` | 12 |
| `proofing` | 9 |
| `settings` | 8 |
| `team` | 8 |
| `templates` | 8 |
| `impersonate` | 7 |
| `common` | 4 |
| `dangerZone` | 4 |
| `gallery` | 4 |
| `account` | 3 |
| `appearance` | 3 |
| `passkeys` | 2 |
| `printShop` | 2 |
| `collections` | 1 |
| `login` | 1 |

So: please rebase onto current `main` and fill the gap. The 397 super-admin keys are the bulk of it. That area is the platform administration interface — it matters more for self-hosters running several studios than it does for us, so it is worth translating properly rather than leaving it to fall back to English.

Two things that should make it easier:

**Use `npm run check:i18n`** (new in `v0.60.0`, `apps/frontend/scripts/check-i18n.mjs`). It reports keys present in one dictionary but not another, so it will list exactly what `it.ts` still needs. It currently only compares `en` and `de` — if you extend it to cover `it` as part of this PR, that would be a welcome addition and I would take it gladly.

**A few translation notes for the super-admin section**, since it is full of operational vocabulary:

- Some values are deliberately untranslated in `de.ts` too, because they are database or config values shown verbatim: `active`, `invited`, `disabled`, `user`, `access`, `system`, `super_admin`, `pg_dump`, `BACKUP_STATUS_PATH`. Leave those as they are.
- `proofing.lightroomHint` mentions Lightroom's colour label names (`Red`/`Yellow`/`Green`) — those are Lightroom's own English label set and must not be translated. The surrounding explanation should be.
- `super.tdImpersonateAs` and `super.pcDrift` use `{user}` / `{studio}` / `{n}` placeholders; word order can move freely around them.

On your note about the VAT copy: your reasoning is right and I would keep it exactly as you have it. Those defaults are German-market facts, not UI language, and a literal translation is the honest rendering.

Take the time you need — there is no deadline on this.

manuzzi:
Thanks for all the comments, I'll pull the newest main branch and integrate the translation there, I'll create a new PR for that

PR #7, opened 2026-08-14
fix(i18n): migrate hardcoded German strings on public/auth pages (1/4)

manuzzi:
## Context

#4 turned out to be much larger than its description suggests: an audit of the whole frontend found roughly 260-280 hardcoded German strings across ~48 files (bypassing the dictionary entirely), 79 hardcoded `de-DE` locale calls for dates/numbers across 34 files, and a static `<html lang="en">` that's never synced with the active locale.

Given the size, I'm splitting this into 4 reviewable PRs instead of one huge diff:
1. **This PR** — public marketing landing page + login/auth flow (highest user-facing impact, smallest surface)
2. Studio interface + shared components
3. Super-admin interface (the largest single chunk — reportedly near-zero `t()` usage in several pages)
4. Locale-aware dates/numbers/currency/sorting + `document.documentElement.lang` sync

## This PR

- `app/page.tsx` — the apex landing page (multi-tenant cloud mode) had its entire hero copy hardcoded in German. It's a Server Component, so `useT()` isn't available; added a small server-side cookie-based lookup instead of pulling client-only i18n into it.
- `app/login/page.tsx` — "Passwort vergessen?" link was hardcoded; reused the existing `forgotPassword.title` key (already identical in both languages) instead of adding a duplicate.
- `app/auth/impersonate-complete/page.tsx` — support-login flow (heading, "session is being created", failure state, back-to-home link) was fully hardcoded; added `impersonate.*` keys.
- `app/auth/reset-password/page.tsx`, `app/g/[slug]/page.tsx`, `app/e/[token]/page.tsx` — Suspense-fallback "Lädt…"/"Wird geladen…" text lived outside the component that already had `t()` in scope; wired up `common.loading`.
- New keys added to **both** `en.ts` and `de.ts` per the issue's technical requirements.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `next build` (production build succeeds, `/` still statically optimized in default single-tenant mode)
- [x] Drove the dev server with Playwright in multi-tenant mode (`NEXT_PUBLIC_DEPLOYMENT_MODE=multi`) with `lumio_locale=en` and `=de`: landing page and login page render fully in the selected language, and the German output is byte-for-byte the same copy as before (no regression to existing German behavior)

Part of #4

### DIFF
--- a/apps/frontend/src/app/auth/impersonate-complete/page.tsx
+++ b/apps/frontend/src/app/auth/impersonate-complete/page.tsx
@@ -37,11 +37,12 @@ export default function ImpersonateCompletePage() {
 }
 
 function LoadingShell() {
+  const t = useT();
   return (
     <div className="min-h-screen flex items-center justify-center bg-surface-canvas">
       <div className="max-w-md w-full mx-4 rounded-lg border border-line-subtle bg-surface-raised p-6 text-center">
-        <h1 className="text-lg font-semibold mb-2">Support-Login</h1>
-        <p className="text-sm text-ink-tertiary">Session wird angelegt…</p>
+        <h1 className="text-lg font-semibold mb-2">{t("impersonate.supportLoginTitle")}</h1>
+        <p className="text-sm text-ink-tertiary">{t("impersonate.creatingSession")}</p>
       </div>
     </div>
   );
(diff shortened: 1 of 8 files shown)


markusthiel:
Thank you for this — and I owe you a straight answer rather than a polite one.

While you were working on this, I was working on the same issue from the other end, and `main` has now moved a long way: the whole frontend has been migrated, the super-admin area included, and `v0.60.0` shipped with it. Your PR is based on `84dc1de`, so rebasing it would mean resolving conflicts against files whose contents have since changed underneath you. That is a lot of work for a result that is already in place, and I am not going to ask you to do it.

So I am closing this and #8 as superseded — not rejected. To be clear about what that means:

**Your audit was more accurate than mine.** You put the scope at roughly 260–280 hardcoded strings across ~48 files, 79 `de-DE` calls across 34 files, and the missing `lang` sync. My own first estimate was far lower, because I was scanning for German-specific characters and that silently skips every string without an umlaut — "Status", "Versendet", "Fortschritt". I corrected the method twice before landing on something reliable. Your number was right from the start.

**And you found real bugs I did not.** The dangling `t()` references you mention in #8 — `studio.cancel`, `studio.save`, `templates.tenantDefault` — are genuine, and when I went looking there were nine of them, not three. Two were inside `confirm()` calls, so the user was being asked "studio.tagConfirmDelete" before deleting a tag. These were invisible to my approach by construction: I was searching for German text, and a `t()` call pointing at nothing contains none. `tsc` cannot see them either, because the dictionary type is `Record<string, string | Dict>`.

They are fixed in `v0.60.0`, and there is now a `npm run check:i18n` that fails the build on dangling references, key drift between the dictionaries, and hardcoded locale identifiers — so this class of bug cannot come back quietly. That check exists because of your report. It is credited in the changelog.

Parts 3 and 4 of your plan (super-admin, locale-aware formatting) are both done, so please do not spend time on those.

If you want to keep contributing, #6 is the thing that matters now and I have left concrete notes there.

PR #8, opened 2026-08-14
fix(i18n): migrate hardcoded German strings in Studio interface (2/4)

manuzzi:
Part 2/4 of #4 (see #7 for 1/4 — public/auth pages). Same approach: literal German text moved into the dictionary, new keys added to **both** `en.ts` and `de.ts`, existing keys reused wherever an exact match already existed instead of adding duplicates.

## Scope

Studio backend pages (galleries list + detail, account, appearance, billing, collections, print shop products/providers/settings, team, templates) and shared components used across them: `NotificationSettings`, `PasskeysSection`, `ProofingPanel`, `SectionsEditor`, `DangerZone`, `MarkdownField`, `SharePanel`, `TwoFactorSection`, `AppearanceFields`, `GalleryHeaderEditor`, `StudioShell` nav, and the gallery `VideoPlayer`. A few of these (`NotificationSettings`, `ProofingPanel`) had essentially zero `t()` usage before this — every visible string was a literal.

## Bonus fixes found along the way

While verifying key coverage I found three **pre-existing** broken references — not introduced by this PR, but real bugs: `SectionsEditor.tsx` called `t("studio.cancel")` / `t("studio.save")` and the template editor called `t("templates.tenantDefault")`. None of those keys exist in either dictionary, so users were seeing the literal key string (`studio.cancel`) rendered as button text, in **any** language. Fixed by pointing at the correct existing keys (`common.cancel`/`common.save`, `studio.brandingTenantDefault`).

Also: `StudioShell`'s nav fallback strings (used only if a dictionary lookup ever fails) were hardcoded in German — inconsistent with the app's English-default fallback chain described in `lib/i18n.tsx`. Switched to English.

## Out of scope here

`de-DE`-hardcoded date/number formatting in these same files (e.g. `DangerZone.tsx`, `PasskeysSection.tsx`) is left as-is — that's part 4/4 along with the `document.lang` sync, since it needs locale-aware `Intl` plumbing rather than a dictionary swap, and mixing the two concerns in one diff seemed harder to review.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `next build` (production build succeeds)
- [x] Every `t()` call touched or added in this diff verified **programmatically** against both `en.ts` and `de.ts` — 805 distinct key paths across the 25 changed files, zero unresolved in either dictionary (catches typos/wrong-section references that `tsc` can't, since `t()` takes a plain string)
- [x] Dev server smoke test in English and German — no console/page errors

Part of #4

### DIFF
--- a/apps/frontend/src/app/studio/[id]/page.tsx
+++ b/apps/frontend/src/app/studio/[id]/page.tsx
@@ -1940,7 +1940,7 @@ export default function GalleryDetailPage() {
                 size="sm"
                 onClick={() => setDupDialog(null)}
               >
-                Abbrechen
+                {t("common.cancel")}
               </Button>
               {dupDialog.newFiles.length > 0 && (
                 <Button
@@ -2019,7 +2019,7 @@ export default function GalleryDetailPage() {
                 onClick={() => setConfirmDialog(null)}
                 disabled={bulkPending}
               >
-                Abbrechen
+                {t("common.cancel")}
               </Button>
               <Button
                 variant={confirmDialog.confirmVariant}
@@ -2073,7 +2073,7 @@ export default function GalleryDetailPage() {
                 size="sm"
                 onClick={() => setLimitDialog(null)}
               >
-                Schließen
+                {t("common.close")}
               </Button>
               <Link
                 href="/studio/billing"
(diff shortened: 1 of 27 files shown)


markusthiel:
Superseded by the work that landed in `v0.60.0` — see my comment on #7 for the full explanation, including why I am closing rather than asking you to rebase.

Specifically on this PR: the three broken `t()` references you found here were real, and looking properly turned up nine. They are fixed, and `npm run check:i18n` now guards against them. That check exists because you reported this.

PR #9, opened 2026-08-14
feat(i18n): add Italian translation

### LINKED ISSUE #5: Italian Translation

manuzzi:
Continues #6, which GitHub won't let me reopen — the fork it was originally submitted from was deleted at some point and GitHub permanently locks the PR's state once that happens ("state cannot be changed. The repository that submitted this pull request has been deleted."), even though the fork exists again now. Same branch, same history, just a fresh PR number. Full prior review context is in #6, in particular [this comment](https://github.com/markusthiel/lumio/pull/6#issuecomment-5295636882).

## Summary
- Adds `apps/frontend/src/lib/i18n/it.ts` — full Italian translation of the frontend dictionary, informal "tu" register throughout, 1:1 against `en.ts` (same key names, nesting, section order).
- Registers the `it` locale in `dict.ts` and `SUPPORTED` in `i18n.tsx`, and the two components carrying the locale union directly (`GalleryShell.tsx`, `app/studio/settings/page.tsx`), including the "Italiano" label in both language pickers.
- Rebased onto current `main` (`v0.63.0`) and filled the gap `main` had accumulated since this branch's original base (`84dc1de`) — 477 keys, almost all in `super` (the platform administration area, translated properly since it matters more for self-hosters than for us).
- Extended `check-i18n.mjs` to compare `it.ts` against `en.ts` as well as `de.ts`, per the maintainer's suggestion in #6.

## Translation notes
- Left untranslated, matching `de.ts`: the literal status values `invited`/`active`/`disabled` (`team.status*`, `super.usersStatus*`), and Lightroom's own `Red`/`Yellow`/`Green` label names in `proofing.lightroomHint`.
- `super.tdImpersonateAs` / `super.pcDrift` placeholders (`{user}`/`{studio}`/`{n}`) translated with free word order around them, as suggested.
- VAT/legal copy (19%/7%, "mandatory in Germany") translated literally rather than re-localized to Italy — those are facts about the product's German origin, not UI language, matching how `de.ts` itself reads.
- The hardcoded German "Passwort vergessen?" in `login/page.tsx` (flagged in the original PR) is still untouched — out of scope for a translation-only PR, tracked under #4.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npm run check:i18n` — 2437 keys in all three dictionaries, no dangling `t()` references, no hardcoded locales
- [x] `next build` — production build succeeds, all routes including every `/super/*` page
- [ ] `next lint` doesn't currently work on `main` (Next 16 removed `next lint`, no standalone eslint config in the repo) — couldn't run it, flagging in case that's news
- [ ] Wasn't able to do a full authenticated browser pass over `/super/*` — no local Postgres/API instance set up in this environment to sign in as a super admin. Everything else (`tsc`, `check:i18n`, `build`) passed, but a native-speaker/visual check of the super-admin pages specifically would still be worth it before merging

Closes #5


markusthiel:
Checked out and verified against current `main`. This is ready — thank you for turning it around so quickly, and for the fresh-PR explanation about the deleted fork; that saved me guessing why #6 wouldn't reopen.

What I checked locally:

- **2437 keys in `it.ts`, 2437 in `en.ts`, zero missing, zero extra.** That includes the ~470-key gap from the rebase and everything `main` picked up today, right up to `v0.63.0`.
- `tsc --noEmit` clean, `next build` clean (54 routes).
- `npm run check:i18n` passes and now reports "all three dictionaries".

Two things you did that I'd have had to ask for otherwise:

- **Extending `check-i18n.mjs` to cover `it`.** That was the right instinct — it also means the dangling-`t()` check now requires a key in all three, so a future key added to only two fails the build. Exactly what I wanted and didn't specify precisely.
- **`format.ts`.** I did not mention it and would probably have missed it in review: `BCP47` maps our locales onto tags for dates, numbers and currency. Without `it: "it-IT"` an Italian interface would have shown US date and currency formatting while every string around it was Italian. Good catch.

Spot-checked the translation notes from my earlier comment and they all held:

- `usersStatusActive` stays `"active"` — the database value, correctly left alone, while `statusInvited` reads `"Invitato"` where it is a display label. That distinction is easy to get wrong in bulk and you got it right.
- `lightroomHint` keeps `Red/Yellow/Green` untranslated while translating the surrounding explanation, and correctly tells the user to set Lightroom to the English label set. That is the substance of the hint, and translating those three words would have broken it.

The register reads consistently informal throughout, which matches how the German and English dictionaries address people.

Merging. If you want to keep going: the API mail templates are now locale-aware end to end (`apps/api/src/services/mail-i18n.ts`), but only carry `de` and `en`. Adding `it` there is a bigger piece — 32 templates, and the `Phrase` type is `Record<MailLocale, string>` on purpose, so adding a locale fails to compile until every template is translated. No pressure at all, and no rush if you do: it is a natural follow-up rather than something missing.

manuzzi:
Added Italian to the mail layer as a follow-up, since it's locale-aware end to end but `MailLocale` only carried `de`/`en`.

- Added `"it"` to `MailLocale` and translated all 33 templates across `mail.ts` and `mail-print.ts`. `Phrase` is `Record<MailLocale, string>` by design, so this was compile-enforced per template — `tsc` wouldn't pass until every one had an `it` entry.
- Fixed the `locale === "de" ? "de-DE" : "en-GB"` date-formatting ternary (13 call sites across `mail.ts`, `mail-print.ts`, `notifier.ts`, `sweeper.ts`) into a `mailBcp47()` helper backed by its own `Record<MailLocale, string>` in `mail-i18n.ts`. Same class of bug as the `format.ts` BCP47 fix you called out on the frontend PR: an unhandled third locale would have silently formatted Italian mail with English dates and currency. Updated `check-mail-i18n.mjs`'s allowlist to recognize the new helper call instead of the old two-locale ternary shape.
- Found two hardcoded German strings (a heading and a button label) in `tmplPrintOrderNotifyStudio`'s HTML body that bypassed `phrase()` entirely — always German regardless of recipient locale. Templated those too while I was in there.

Test plan:
- [x] `npx tsc --noEmit` — clean
- [x] `npm run check:mail-i18n` — passes, no hardcoded locales or untemplated mails
- [x] `npm run build` — clean
- [x] `npx vitest run` — 159/159 passing

markusthiel:
Merged — your commits `e2b3cce` and `9fbd775` are on `main` as of `4e956ab`, unchanged and fast-forwarded, so the history is exactly what you pushed.

GitHub will leave this PR showing "open" rather than "merged", and that is not a mistake on either side: this repository is a push mirror of a self-hosted Forgejo instance, which is the source of truth. Merging through the GitHub button would have put a merge commit on a branch that the next mirror sync overwrites — so the merge happened upstream and came back down. GitHub has no way to recognise that, so I am closing this by hand. Your commits are in, with authorship intact.

The only thing I added on top was a CHANGELOG entry, crediting you. You were right not to touch it — a contributor shouldn't have to guess our release wording.

Italian ships in the next release.

Issue #12, opened 2026-08-19
Findings from running Lumio self-hosted with a third UI language — Lightroom plug-in, locale drift, and hardcoded German strings


canja006:
Hi, and thanks for Lumio.

I run a self-hosted instance for my photography business (real estate and holiday cottages, Finland). Over the past few days I got the whole workflow going end to end — publish from Lightroom Classic, client selection, selection import back into the catalog — and added Finnish as a third UI language.

CONTRIBUTING.md asks for an issue before larger changes, so here is what I found, before I open any pull requests. Everything below is on v0.64.0.

I'd rather ask than assume, so there are a couple of questions at the end.

## 1. The Lightroom plug-in does not load

`apps/lightroom-plugin` fails at load time: `LrBinding.makePropertyTable(_propertyTable.context)` throws because LR does not expose a `.context` field on the property table. That surfaces as *"Could not create info sections for plug-in"*, the plug-in is flagged broken, and none of its menu items register. `viewForCollectionSettings` also returns a sections table where LR expects a single view, and `getCollectionBehaviorInfo` is missing, so creating a collection fails with `attempt to call method 'hierarchyCreated' (a nil value)`.

Once it loads, two things fail silently rather than loudly:

- **Selection import matches nothing.** Publishing forces the uploaded name to `<basename>.jpg`, but import indexes the catalog by its real filename including the extension, so `int_4325.jpg` can never match `int_4325.nef`. On my catalog (dng 1083, nef 967, jpg 91, tif 7) that is 95.8 % of photos that could never be found.
- **Metadata writes fail without reporting it.** The writes are wrapped in a plain `pcall`, but `photo:getRawMetadata` / `setRawMetadata` are yielding catalog operations, so every write fails with *"Yielding is not allowed within a C or metamethod call"* while the summary still reports success. My first run said "57 photos updated" and had written nothing.

Eleven fixes in total, including `recordPublishedPhotoUrl(nil)` (LR requires a string, so every photo failed and none were marked published — they would be re-uploaded on every publish), retry handling that is careful not to repeat non-idempotent POSTs, and `rendition:uploadFailed()` on the failure paths.

## 2. Locale handling drifts because the same list is written in several places

Three independent instances of the same shape, all of which work by accident for exactly two languages:

[...]

PR #13, opened 2026-08-19
fix(lightroom-plugin): repair plugin loading, uploads and selection import

canja006:
Fixes the Lightroom Classic plug-in, which does not load against a self-hosted instance and, once loaded, fails silently in two ways. Part of #12 — this is the independent piece; the locale work is separate.

Tested end to end against v0.64.0 with Lightroom Classic on macOS and a real catalog (dng 1083, nef 967, jpg 91, tif 7). Rebased onto current `main` (v0.66.3); touches only the six `.lua` files.

### Blocking — the plug-in did not load
- **PluginManager**: `LrBinding.makePropertyTable(_propertyTable.context)` throws because LR does not expose `.context` → *"Could not create info sections for plug-in"* → no menu items register. LR already hands over a bindable table.
- **LumioPublishService**: `viewForCollectionSettings` returned a sections table where LR expects a single view; `getCollectionBehaviorInfo` was missing entirely. Both surface as `attempt to call method 'hierarchyCreated' (a nil value)`.

### Silent data loss — appeared to work, wrote nothing
- **ImportSelectionTask**: a plain `pcall` wrapped the metadata writes, but `photo:getRawMetadata`/`setRawMetadata` yield, so every write failed with *"Yielding is not allowed within a C or metamethod call"* while the summary reported success. My first run said "57 photos updated" and had written nothing. Now `LrTasks.pcall`, with failures counted and logged.
- **ImportSelectionTask**: matching compared the catalog filename *with* extension against the name publishing sends, which is always `<basename>.jpg`, so `int_4325.jpg` could never match `int_4325.nef` — 95.8 % of the test catalog. Now exact name first, then basename, preferring the RAW/DNG/TIF master over an exported JPEG. Ambiguous basename matches are skipped and reported rather than written to all.
- **LumioPublishService**: `recordPublishedPhotoUrl(nil)` failed every photo with *"URL must be a string"*; uploads succeeded but LR never marked them published, so they would re-upload on every publish.

### Robustness
- **LumioApi**: retry with backoff for transient errors, but only for GET and for 429 on POST — `/uploads/init` and `/plugin/galleries` are not idempotent, so blind POST retries risked duplicate galleries and orphan rows. The S3 PUT is retried (idempotent) and now requires an explicit 2xx; a missing status is an error rather than a silent nil.
- **LumioPublishService**: `rendition:uploadFailed()` on both failure paths, and a failing `status=live` PATCH is now visible instead of leaving the gallery a draft with no warning.

The plug-in UI is translated from German to English per the language policy, and the new comments are in English.

### DIFF
--- a/apps/lightroom-plugin/lumio.lrdevplugin/ImportSelectionDialog.lua
+++ b/apps/lightroom-plugin/lumio.lrdevplugin/ImportSelectionDialog.lua
@@ -25,10 +25,10 @@ LrTasks.startAsyncTask(function()
     LrFunctionContext.callWithContext("ImportSelectionDialog", function(context)
 
         -- Galerien laden, mit User-friendly Error
-        local ok, galleries = pcall(LumioApi.listGalleries)
+        local ok, galleries = LrTasks.pcall(LumioApi.listGalleries)
         if not ok then
             LrDialogs.message(
-                "Lumio: Verbindung fehlgeschlagen",
+                "Lumio: connection failed",
                 tostring(galleries):gsub("^.-: ", ""),
                 "critical"
             )
@@ -37,8 +37,8 @@ LrTasks.startAsyncTask(function()
         if #galleries == 0 then
             LrDialogs.message(
                 "Lumio",
-                "Keine Galerien gefunden. Lege im Studio eine an oder prüfe, " ..
-                "dass der Token zum richtigen User gehört."
+                "No galleries found. Create one in the studio, or check " ..
+                "that the token belongs to the right user."
             )
             return
         end
@@ -71,7 +71,7 @@ LrTasks.startAsyncTask(function()
 
             f:row {
                 f:static_text {
-                    title = "Galerie:",
+                    title = "Gallery:",
                     width = 100,
                 },
                 f:popup_menu {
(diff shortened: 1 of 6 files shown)


markusthiel:
Merged — `7ecc8d8` is on `main`, fast-forwarded, so your commit is unchanged.

Thank you for this. The plug-in has never had a proper test against a real RAW catalog, and it shows: three of these are failures that report success, which is the worst kind to have shipped.

What I verified on my side, and what I could not:

- **Verified:** all six files are syntactically valid Lua (`luac -p`; I also confirmed the check itself catches a deliberate error, so it means something). The diff touches nothing outside `apps/lightroom-plugin`. Every endpoint `LumioApi.lua` calls exists server-side — `/plugin/version`, `/plugin/galleries`, `/uploads/init`, `/uploads/complete`, `/plugin/galleries/:id/selection`. And the mismatch at the heart of this is real in the pre-merge code: `LumioPublishService` built `nameNoExt .. ".jpg"` while `ImportSelectionTask` indexed on `getFormattedMetadata("fileName")`. Those two can never agree for a RAW file.
- **Not verified:** the runtime behaviour in Lightroom. I have no catalog and no LR install, so the SDK-level claims — `.context` not existing, the yielding metadata calls, `recordPublishedPhotoUrl` rejecting nil — rest entirely on your testing. Your numbers (dng 1083, nef 967, jpg 91, tif 7) made that easy to trust, and the reasoning is specific enough that I would rather take it than pretend to check it.

Two decisions in here I want to name, because they are better than what I would have done:

Skipping ambiguous basename matches instead of writing to all of them. The tempting fix is to write to every candidate and move on; you chose to skip and report. A silent multi-write into someone's catalog is exactly the kind of thing that gets discovered weeks later.

Retrying GET but not POST, except for 429. `/uploads/init` and `/plugin/galleries` are not idempotent and a blind retry would leave duplicate galleries and orphan rows behind. That distinction is easy to skip when you are adding "just some retry logic".

The smoke test is green on the merge commit. Released shortly.

PR 2 whenever suits you — no hurry. The tests around the locale helpers are the part I am most looking forward to.

PR #14, opened 2026-08-20
fix(mail): localise the layout footer + test the locale helpers

canja006:
The remaining piece of #12 after your v0.66.3–v0.67 work: the mail-layout footer, and tests for the locale helpers. The three drift fixes are all upstream now, so this is only those two. Rebased onto current `main` (v0.69.0); touches `mail-layout.ts` and adds `mail-i18n.test.ts`, nothing else.

### Footer
It was hardcoded German and claimed the service is *"gehostet in Deutschland"*, which is untrue on a self-hosted instance. It is now a `Phrase` and follows the recipient's locale, so the type demands a translation per mail locale. `RenderMailOpts` gains an optional `locale`; callers that resolved one pass it, otherwise the instance default applies — no change for existing deployments. The link points at the repository rather than `lumio-cloud.de`, per your call in #12.

Renders as:
- `de` — `Verschickt von <a …>Lumio</a>`
- `en` — `Sent via <a …>Lumio</a>`

### Tests
Eleven tests for `localeTag`, `normalizeLocale`, `phrase` and `interpolate` — there were none. The formatting assertions guard the regression from #12: with a hardcoded `de-DE : en-GB` ternary, any non-German locale silently formatted as British English. Each locale is asserted to have its own BCP-47 tag and to format dates and currency in its own convention, so the rule against hardcoded locale identifiers becomes something the suite enforces rather than a line in the docs.

Verified against v0.69.0: `tsc --noEmit` clean, all 11 tests pass.

### DIFF
--- a/apps/api/src/services/mail-i18n.test.ts
+++ b/apps/api/src/services/mail-i18n.test.ts
@@ -1,0 +1,108 @@
+/**
+ * Tests for the mail locale helpers.
+ *
+ * The formatting assertions guard a class of bug the codebase was carrying:
+ * dates and currency were formatted with `l === "de" ? "de-DE" : "en-GB"`,
+ * repeated in twelve places. That works by accident for exactly two locales
+ * and silently mis-formats every further one, since anything that is not
+ * German falls through to British English. CONTRIBUTING.md forbids hardcoding
+ * a locale identifier in Intl.*; these tests make the ban enforceable instead
+ * of a convention.
+ */
+import { describe, it, expect } from "vitest";
+import {
+  localeTag,
+  normalizeLocale,
+  phrase,
+  interpolate,
+  type MailLocale,
+  type Phrase,
+} from "./mail-i18n.js";
+
+// Kept in sync with MailLocale by the compiler: adding a locale to the type
+// without adding it here makes this array fail to satisfy MailLocale[].
+const LOCALES: MailLocale[] = ["de", "en"];
+
+const LONG_DATE = {
+  day: "2-digit",
+  month: "long",
+  year: "numeric",
+  timeZone: "UTC",
+} as const;
+
+describe("localeTag", () => {
+  it("returns a BCP-47 tag for every supported locale", () => {
+    for (const l of LOCALES) {
+      expect(localeTag(l)).toMatch(/^[a-z]{2}-[A-Z]{2}$/);
+    }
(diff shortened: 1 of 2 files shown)


markusthiel:
Merged and released in `v0.70.0`. Your commits `c95c97f` and `1614096` are on `main` unchanged.

The tests are the part I am most glad to have. I checked they earn their keep the only way that means anything: I put the #12 ternary back into `localeTag` and ran them — three fail, including both formatting assertions. So they genuinely guard the regression rather than just describing it.

Asserting on behaviour rather than return values is what makes that work. "A German date has a dot after the day number, a British one does not" stays true if we ever change the tag table; `expect(localeTag("de")).toBe("de-DE")` would not have. And typing `LOCALES` as `MailLocale[]` means adding a locale to the type without adding it there fails to compile, so the suite cannot quietly fall behind.

I extended the footer rather than taking it as-is, and it is worth saying why. Your version is right for self-hosting, but it drops the tagline everywhere — including on the hosted instance, where "hosted in Germany" is both true and a selling point. So there are now two optional settings: set `MAIL_FOOTER_TEXT` / `MAIL_FOOTER_URL` and the operator's text wins verbatim; leave them unset and your translated default applies. Existing installs see exactly your change and nothing else.

The tradeoff is documented in `.env.example` rather than left to be discovered: custom text is not translated, so it appears in every mail regardless of the recipient's language. On a multi-language instance the default is the better choice.

I also added a render-level test. Your tests prove the helpers are correct; the bug lived in the layout. It asserts on the produced HTML that the German-hosting claim and the `lumio-cloud.de` link are gone in both locales — closing the gap between "the helper returns the right tag" and "the mail actually says the right thing".

One thing I broke and you should know about, since it was in the part I added on top of your work: I put `MAIL_FOOTER_URL=${MAIL_FOOTER_URL:-}` in docker-compose so Docker would forward it, and `${VAR:-}` puts an *empty string* in the environment when the variable is unset. `z.string().url().optional()` rejects `""` — so the API refused to boot on any instance that does not set the variable, which is all of them. The smoke test caught it; tsc, both i18n checks and all 16 tests were green, because none of them starts the API with a realistic environment. Fixed in `376e2b9` with four tests covering the empty-string case.

PR 3 whenever you are ready. Same as before — no hurry.

PR #15, opened 2026-08-20
fix(mail): translate the layout footer and correct its self-hosted claim

manuzzi:
## Summary

- The mail footer ("Verschickt von Lumio ... gehostet in Deutschland") was a raw German literal in `mail-layout.ts`, appearing on every outgoing email regardless of the recipient's locale — the last hardcoded string in an otherwise fully templated mail layer. It also claimed the service is hosted in Germany (false for a self-hosted instance) and linked to `lumio-cloud.de` instead of the project.
- Turned the footer into a `Phrase` (de/en, matching `MAIL_LOCALES`), pointed the link at the repository, dropped the hosting claim, and made `<html lang>` follow the actual locale instead of always `"de"`.
- `RenderMailOpts.locale` is now required, so a future template can't omit it silently. All 33 `renderMailLayout()` call sites across `mail.ts`, `mail-print.ts` and `broadcast.ts` now pass their resolved locale — from the recipient where one exists (broadcast audience members via `User.locale`, the direct admin-to-user email), and from the instance default where there is no specific recipient (broadcast preview/test-send/archival) or the content is inherently untranslated internal tooling (the support-request notification to the Lumio team, now explicitly `locale: "de"` with a comment explaining why).
- Added the locale-helper unit tests flagged as missing in #12 — `localeTag()` and `normalizeLocale()` had no coverage at all, which is how they broke twice in two days per the issue thread.

Refs #12 — this was the "PR 2" footer + tests work still open after the locale-drift fixes landed.

## Test plan

- [x] `node apps/frontend/scripts/check-i18n.mjs` — passes
- [x] `npx vitest run` in `apps/api` — 170/170 tests pass (11 new)

### DIFF
--- a/apps/api/src/routes/broadcasts.ts
+++ b/apps/api/src/routes/broadcasts.ts
@@ -26,6 +26,7 @@ import {
   type Audience,
 } from "../services/broadcast.js";
 import { sendMail } from "../services/mail.js";
+import { instanceMailLocale } from "../services/mail-i18n.js";
 import { config } from "../config.js";
 
 const audienceSchema = z.enum([
@@ -143,7 +144,8 @@ export async function registerBroadcastRoutes(app: FastifyInstance) {
     // wird in processBroadcast() pro Empfaenger frisch gerendert.
     const archivalHtml = renderBroadcastHtml(
       body.bodyMarkdown,
-      "{{UNSUBSCRIBE_URL}}"
+      "{{UNSUBSCRIBE_URL}}",
+      instanceMailLocale()
     );
 
     const row = await prisma.broadcast.create({
@@ -197,7 +199,13 @@ export async function registerBroadcastRoutes(app: FastifyInstance) {
     req.requireSuperAdmin();
     const body = previewSchema.parse(req.body);
     const exampleUnsubUrl = `${config.PUBLIC_URL.replace(/\/+$/, "")}/broadcasts/unsubscribe?t=example`;
-    return { html: renderBroadcastHtml(body.bodyMarkdown, exampleUnsubUrl) };
+    return {
+      html: renderBroadcastHtml(
+        body.bodyMarkdown,
+        exampleUnsubUrl,
+        instanceMailLocale()
+      ),
+    };
   });
 
   // -------------------------------------------------------------------------
@@ -214,7 +222,11 @@ export async function registerBroadcastRoutes(app: FastifyInstance) {
     const body = testSendSchema.parse(req.body);
 
     const dummyUnsub = `${config.PUBLIC_URL.replace(/\/+$/, "")}/broadcasts/unsubscribe?t=test-only-no-effect`;
-    const html = renderBroadcastHtml(body.bodyMarkdown, dummyUnsub);
+    const html = renderBroadcastHtml(
+      body.bodyMarkdown,
+      dummyUnsub,
+      instanceMailLocale()
+    );
 
     await sendMail({
       to: sa.admin.email,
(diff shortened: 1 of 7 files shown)


canja006:
Ah — we wrote this in parallel and mine happened to land a few minutes before you opened this. Sorry; there was no way for either of us to see the other coming. For what it's worth, yours is the more thorough take, and I'd hate to see the better half of it lost to the timing.

The overlap with what merged (#14, `8b24f84`) is the footer itself — Phrase, repo link, hosting claim dropped — and the eleven locale-helper tests. Those are in `main` now.

But three things in here are **not** in `main`, and they're the part that actually matters:

- **`<html lang>` still hardcoded to `"de"`** on `main` (mail-layout.ts:199). Yours makes it follow the locale.
- **The call sites still don't pass a locale.** On `main`, `mail.ts` has 30 `renderMailLayout()` calls and 0 pass one; `broadcast.ts` 2 and 0. So they all fall back to the instance default — which means a recipient with `User.locale = "en"` on a German-default instance gets German. Your wiring of all 36 call sites to the resolved recipient locale is the difference between "everything in the instance language" and "each in the recipient's". That's the real fix; the footer was just the visible symptom.
- **`RenderMailOpts.locale` required** rather than optional, so a future template can't drop it silently.

If Markus agrees, rebasing this down to just those — the call-site wiring, the `lang` attribute, and the required field — would land the genuinely additive part cleanly on top of what's already there, without the now-redundant footer/tests diff to review. Your call and his, of course.

Either way, thanks for going deeper than I did on this.

markusthiel:
Merged as `759ce4c` — your commit `21319cf` is in the history with authorship intact.

@canja006 was right that the overlap should not cost you the rest of it, and he was right about which part mattered. I reduced the merge to exactly what he suggested: the call-site wiring, the `lang` attribute, and the required `locale` field.

The call sites are the real find here, and I want to be explicit that neither he nor I had seen it. The footer was the visible symptom; the cause was that **none** of the 35 `renderMailLayout()` sites passed a locale, so every mail rendered its frame in the instance default no matter who it went to. English content, German footer, `lang="de"`. I wrote the failing test before merging to make sure I understood the claim rather than taking it on trust — it is in the tree now as `mail-locale-endtoend.test.ts`, so the regression cannot come back quietly.

Making `RenderMailOpts.locale` required rather than optional is the part I would have got wrong. An optional field with a sensible default reads as the safe choice and is precisely how this stayed invisible: 35 call sites all silently taking the fallback, and nothing to notice. Requiring it is what turned the two `broadcast.ts` sites up.

The `lang` attribute is worth more than it looks, too — screen readers take pronunciation from it and spam filters weigh it, so it was wrong in a way that does not show up in a screenshot.

What I did not take, so you are not surprised reading `main`:

- **The footer phrase.** It became configurable in `v0.70.0`/`v0.71.0` (env vars, and a field in the admin UI under System) so a self-hosted instance can carry its own studio name instead of ours. Your `footerBrandPhrase` is redundant against that path, so it is dropped; the translated default it replaced is still there for instances that configure nothing.
- **The helper tests**, since @canja006's landed in #14 first. Not a quality judgement with one exception worth naming: his assert on behaviour ("a German date has a dot after the day number") rather than on the value, which survives a change to the tag table where `expect(localeTag("de")).toBe("de-DE")` would need editing. Both are good tests; that one property made the difference.

Answered #16 as well — yes to Italian in `MAIL_LOCALES`, one PR, `DEFAULT_MAIL_LOCALE` staying `de`. Rebase onto current `main` first; all three mail files have moved under you.

Issue #16, opened 2026-08-20
Extend MAIL_LOCALES to include Italian


manuzzi:
Follow-up to #12.

## Context

The frontend already treats Italian as a first-class locale: `Locale = "en" | "de" | "it"` in `dict.ts`, with a complete `it.ts` dictionary (2560 keys). The mail layer stops at German/English: `MAIL_LOCALES = ["de", "en"] as const` in `mail-i18n.ts`, and `Phrase = Record<MailLocale, string>` enforces both by design.

I ran into the gap directly while closing the last hardcoded-German string from #12 (the mail-layout footer — PR incoming separately). The footer needed a `de`/`en` Phrase like everything else, which was a reminder that Italian mail recipients get English, never Italian, for anything the mail layer sends.

## Scope, measured not estimated

- `mail.ts`: 174 Phrase entries across ~29 `tmpl*` functions
- `mail-print.ts`: 8 more
- **~182 phrases total**, each needing a reviewed Italian value before `tsc` passes — `Phrase`'s all-or-nothing typing means this isn't incremental

Same order of magnitude as the Finnish mail work (335 phrases) referenced in #12.

## What I'm proposing

- Add `"it"` to `MAIL_LOCALES`
- Translate all ~182 phrases myself (I maintain the Italian UI locale; per #12 I understand machine-translated Italian without native review isn't wanted)
- Leave `DEFAULT_MAIL_LOCALE` on `de` — no behavior change for existing instances
- Branch: `feat/mail-locale-it`, opened once there's a green light

## Before I start translating ~180 phrases

Per CONTRIBUTING.md and what #12 already showed about translation work done before agreement on scope:

1. Is extending `MAIL_LOCALES` wanted, or is `it` deliberately mail-scoped to de/en for a reason I'm not seeing?
2. `DEFAULT_MAIL_LOCALE` staying `de` — correct?
3. Split `mail.ts`/`mail-print.ts` into separate PRs given the size, or one PR?

manuzzi:
Quick update while this waits for your read.

[...]

PR #17, opened 2026-08-21
feat(mail): extend MAIL_LOCALES with Italian

manuzzi:
## Summary

- Adds `"it"` to `MAIL_LOCALES` in `mail-i18n.ts` — the single source of truth for supported mail languages. `Phrase = Record<MailLocale, string>` is all-or-nothing by design, so this required a reviewed Italian value for every mail template: 307 phrases in `mail.ts`, 33 in `mail-print.ts`, 3 in the shared `common` block, plus `localeTag()`'s BCP-47 mapping (`it` → `it-IT`). All translated natively — no machine translation, per the standard set in #12.
- Frontend: the `"de" | "en"` unions for the personal mail-locale endpoint and the tenant customer-mail-locale setting now include `"it"` end to end (`api.ts`, `studio/settings/page.tsx`). Removed the write-through special-case that skipped `/settings/my-locale` for `"it"` — it existed only because the server used to reject that value, and it doesn't anymore.
- `DEFAULT_MAIL_LOCALE` stays `de` — no behavior change for instances that configure nothing, same guarantee given for prior locale additions. Its zod schema is a literal list (kept manually in sync with `MAIL_LOCALES`, documented in place) rather than an import, since importing from `mail-i18n.ts` would make it circular with `config.ts`.
- Two hardcoded German strings found in `mail-print.ts` while working through its templates (a heading and a button label that never went through `phrase()` for *any* language, not just Italian) — translated alongside the Italian work since they were sitting right there.

Rebased onto current `main` after #15 landed and the mail layer moved further (configurable footer, required `locale` on every `renderMailLayout()` call). Two things fell out of that rebase, not just conflict resolution:

- `mail-layout.ts` grew its own `layoutPhrases.sentVia` Phrase for the new configurable-footer default text — needed its Italian value too.
- `mail-i18n.test.ts`'s `LOCALES` array was a hand-written `["de", "en"]` with a comment claiming the compiler keeps it in sync with `MailLocale` — not true for a plain array type, so it was silently only testing two locales. Now imports `MAIL_LOCALES` directly (same fix in `mail-footer.render.test.ts`), so the Finnish locale lands with test coverage for free instead of needing its own follow-up patch to these files.

Refs #12, #16

## Test plan

- [x] `node apps/frontend/scripts/check-i18n.mjs` — passes (2567 keys)
- [x] `npx vitest run` in `apps/api` — 184/184 tests pass across 23 files, including the new `mail-locale-endtoend.test.ts`, `mail-footer.render.test.ts`, `mail-footer-precedence.test.ts`, `config.footer.test.ts`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ These stay German-first, on purpose:
   of a studio, not the studio operator, so their language follows the studio
   rather than our default. That is now implemented: `mail-i18n.ts` resolves
   `Tenant.locale` for customer mail and `User.locale` for team mail, and every
-  template carries a `de`/`en` pair. What stays deliberate is that
+  template carries a `de`/`en`/`it` set. What stays deliberate is that
   `DEFAULT_MAIL_LOCALE` defaults to `de`, so an instance that sets nothing
   keeps writing German.
 
--- a/apps/api/src/config.ts
+++ b/apps/api/src/config.ts
@@ -99,7 +99,10 @@ const envSchema = z.object({
   // Cookie des jeweiligen Browsers und wird hiervon nicht beeinflusst.
   // Optional mit Default, also kein Breaking Change fuer Self-Hoster:
   // ohne Eintrag bleibt alles auf Deutsch wie bisher.
-  DEFAULT_MAIL_LOCALE: z.enum(["de", "en"]).default("de"),
+  // Literal-Liste statt Import aus mail-i18n.ts (das wuerde config.ts
+  // <-> mail-i18n.ts zirkulaer machen, mail-i18n.ts importiert config
+  // bereits) — MUSS mit MAIL_LOCALES dort synchron gehalten werden.
+  DEFAULT_MAIL_LOCALE: z.enum(["de", "en", "it"]).default("de"),
 
   /**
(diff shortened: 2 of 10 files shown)


markusthiel:
Merged and released in `v0.72.0`. Thank you — 338 phrases with a reviewed value each is a lot of careful work.

I spot-checked the deletion and billing deadlines specifically, since those carry commitments: "60 giorni in un periodo di tolleranza", "cancellati definitivamente". Nothing softened. An Italian welcome mail comes out Italian throughout, including `lang="it"` and "01 settembre 2026" rather than a German date.

Two of your findings outlive the translation. The test file only ever exercised two locales — and the comment above it claimed the compiler kept the array in sync with `MailLocale`. It does not; a subset satisfies `MailLocale[]` happily. That comment was mine, and it made a test look like a guarantee for months. Importing `MAIL_LOCALES` directly is the right fix, and it means Finnish arrives with coverage instead of needing a follow-up.

The two hardcoded strings in `mail-print.ts` were invisible to `check:mail-i18n` for the same kind of reason: it looks for locale identifiers and untemplated sends, not for a phrase that was never created.

@canja006 — Italian is in, so Finnish is clear to go on top of three existing columns as you planned.

PR #20, opened 2026-08-21
feat(mail): add Finnish as a mail locale

canja006:
Adds Finnish as a mail locale, following @manuzzi's Italian work in #17 — `fi` now sits alongside `de`/`en`/`it` across the whole mail layer. Part of #12; Markus gave the go-ahead for Finnish on top of the three existing columns once Italian landed.

Rebased onto current `main`. 340 phrases, each with a reviewed value: 307 in `mail.ts`, 33 in `mail-print.ts`, plus the shared `common` phrases and the layout footer.

### What's covered
- `MAIL_LOCALES` gains `"fi"`; `MailLocale`, the config schema and both request validations derive from it, so nothing drifts.
- `localeTag` gains `fi-FI`, so dates and currency format in Finnish convention (`30. syyskuuta 2026`, decimal comma) rather than falling through to English.
- The footer `sentVia` phrase and the shared `common` phrases get their `fi` value.
- Because `Phrase` is `Record<MailLocale, string>` and the tests import `MAIL_LOCALES` directly, adding `"fi"` made the compiler demand a Finnish value in every template and the suite exercise every locale — Finnish arrives with coverage, no follow-up needed.

### Translation notes
Terminology follows the same glossary used across the app, which matters where Finnish is ambiguous: upload = *lähetä* vs download = *lataa* (Finnish "lataa" means both), tag = *tunniste* (not *avainsana*, which collides with Lightroom keyword metadata), tenant = *studio*. Finnish quotation marks `” ”` throughout.

### Scope
Mail layer only. `DEFAULT_MAIL_LOCALE` stays `"de"`, so existing instances see no change. The studio settings **UI** language picker stays `de`/`en`/`it` — a Finnish UI locale (`fi.ts`) is a separate change; here `fi` is offered only in the **tenant mail-language** selector, which is purely about which language a studio's customers are written in.

### Verified against current `main`
- [x] `tsc --noEmit` clean in `apps/api` and `apps/frontend`
- [x] `npx vitest run` — 184/184 tests pass


markusthiel:
Both merged and released in `v0.73.0`. Thank you — 2907 strings between the two PRs, each with a reviewed value.

I checked the Finnish formats against rendered output rather than the description: `30. syyskuuta 2026` for dates, `1 234,50 €` for currency, `lang="fi"` on the mail frame. All as you said.

The terminology notes are the part I could not have checked and would not have known to ask about. `lähetä` for upload against `lataa` for download is the kind of decision that is invisible when it is right and everywhere when it is wrong, and `tunniste` over `avainsana` because of the Lightroom collision is the same care you showed in the plug-in PR.

One thing on my side that your work exposed: `check-i18n` was not checking `fi.ts` at all. It read `de.ts` and `it.ts` by name, three hardcoded places each, and went on reporting "all three dictionaries" — an incomplete Finnish file would have passed in silence. Same failure as the one @manuzzi found in the mail tests: a check that promises completeness and quietly stops delivering it when a language is added. The locale list now comes from `dict.ts`.

Two things fixed after release, both from screenshots of the Finnish UI:

The language setting read `Language / Sprache` with an English-only hint — my comment in that file argued German helped someone stranded in an unreadable language, which was true with two languages and stopped being true at four. It now reads `Language / Kieli` with the hint in both.

The gallery statistics funnel was still German: the heading was hardcoded, and the four step labels come from the API as prose next to a stable key. That is the third time that arrangement has shipped untranslated, so the check now scans the API for the *shape* rather than a list of known files. It immediately found the print provider config fields too.

PR #21, opened 2026-08-21
feat(i18n): add Finnish UI translation

canja006:
Adds Finnish as a frontend UI locale — `fi.ts` with all 2567 keys, registered the same way as @manuzzi's Italian. Completes the Finnish support started with the mail locale in #20; part of #12.

The dictionary is complete: `LocaleDict` (= `typeof en`) makes `tsc` reject any missing or extra key, and `check-i18n` confirms 2567 keys across all locales. Nothing falls back to English.

### Registered in
- `dict.ts` — import, `Locale` union, `dictionaries` map
- `i18n.tsx` — `SUPPORTED`, so cookie / `navigator.language` detection picks Finnish up
- `format.ts` — `fi-FI`, so dates, numbers, currency and sorting follow Finnish convention
- `GalleryShell.tsx` — the client-facing gallery language picker
- `settings/page.tsx` — the studio UI language picker

### Translation notes
Terminology matches the mail work and the shared glossary, which matters where Finnish is ambiguous: upload = *lähetä* vs download = *lataa* (Finnish "lataa" is both), tag = *tunniste* (not *avainsana*, which collides with Lightroom keyword metadata), tenant = *studio*, proofing = *kuvavalinta*. Finnish quotation marks `” ”` throughout. The new `apiError.*` section is fully translated.

### Depends on #20
The settings language picker writes through to `updateMyMailLocale`, so `api.ts` needs `fi` in its mail-locale type. That one change is included here and is **identical to #20** — it merges cleanly whichever lands first. If you merge #20 first, that hunk simply disappears on rebase.

### Verified against current `main`
- [x] `tsc --noEmit` clean
- [x] `npm run build` succeeds
- [x] `node scripts/check-i18n.mjs` — 2567 keys, no dangling `t()` refs, no hardcoded locales


markusthiel:
Both merged and released in `v0.73.0`. Thank you — 2907 strings between the two PRs, each with a reviewed value.

I checked the Finnish formats against rendered output rather than the description: `30. syyskuuta 2026` for dates, `1 234,50 €` for currency, `lang="fi"` on the mail frame. All as you said.

The terminology notes are the part I could not have checked and would not have known to ask about. `lähetä` for upload against `lataa` for download is the kind of decision that is invisible when it is right and everywhere when it is wrong, and `tunniste` over `avainsana` because of the Lightroom collision is the same care you showed in the plug-in PR.

One thing on my side that your work exposed: `check-i18n` was not checking `fi.ts` at all. It read `de.ts` and `it.ts` by name, three hardcoded places each, and went on reporting "all three dictionaries" — an incomplete Finnish file would have passed in silence. Same failure as the one @manuzzi found in the mail tests: a check that promises completeness and quietly stops delivering it when a language is added. The locale list now comes from `dict.ts`.

Two things fixed after release, both from screenshots of the Finnish UI:

The language setting read `Language / Sprache` with an English-only hint — my comment in that file argued German helped someone stranded in an unreadable language, which was true with two languages and stopped being true at four. It now reads `Language / Kieli` with the hint in both.

The gallery statistics funnel was still German: the heading was hardcoded, and the four step labels come from the API as prose next to a stable key. That is the third time that arrangement has shipped untranslated, so the check now scans the API for the *shape* rather than a list of known files. It immediately found the print provider config fields too.

PR #25, opened 2026-08-31
feat(gallery): permanent delete for archived galleries

### LINKED ISSUE #24: Galleries can be archived but never permanently deleted
## Problem

The studio dashboard's gallery card menu ("⋮") only offers "Archive" /
"Reactivate" — there's no way to permanently delete a gallery. Archived
galleries just accumulate forever.

This looks half-shipped rather than missing: the backend already has a working
`DELETE /galleries/:id` (hard delete, cascades to files/share links/comments,
audit-logged, triggers async S3 cleanup) and the frontend API client already has
`api.deleteGallery(id)` wired to it — nothing in the UI ever calls it.

Two real gaps need fixing in the backend endpoint itself before it's safe to
expose in the UI:

- Nothing stops it from deleting a gallery that has print orders, which would
  destroy financial/order records (payment refs, amounts, shipping address).
- Nothing restricts *who* can delete — today "access" (view/edit/delete/share)
  is a single undifferentiated right; permanent deletion needs its own, stricter
  rule.

## Proposed solution

- Add "Delete permanently" to the dashboard gallery card menu, visible only once
  a gallery is archived (never allow hard-deleting a `live` gallery a client
  might currently be viewing).
- Require typing the gallery's title to confirm (compared case-insensitively,
  matching the existing `DangerZone.tsx` pattern).
- Server-side guard: 409 if not archived, 409 if the gallery has any print
  orders — enforced in the API itself (also used directly by the Lightroom
  plugin/scripts), not just the UI.
- Server-side role guard: Owners and Admins can delete; Members never can.
  (Corrected 2026-09-09 — this originally had it the other way round, with
  Admins above Owners. In Lumio `owner` is the studio owner and the top role;
  `galleryAccessWhere()` already scopes what an admin can reach before the
  delete check runs, so the rule is simply "not a member" and needs no
  per-gallery ownership test.)

## Acceptance criteria

- [ ] `DELETE /galleries/:id` returns 409 if the gallery is not archived.
- [ ] `DELETE /galleries/:id` returns 409 if the gallery has any print orders.
- [ ] `DELETE /galleries/:id` returns 403 if the caller is a Member. Owners and
      Admins are both allowed; reachability is already handled by
      `galleryAccessWhere()`, which 404s before the role check.
- [ ] Dashboard gallery card menu has "Delete permanently", shown only for
      archived galleries the current user is allowed to delete.
- [ ] Deleting requires typing the gallery's title, compared
      case-insensitively.
- [ ] New error cases show a friendly, translated message.
- [ ] Existing delete behavior (cascade, audit log, `gallery.deleted` event, S3
      cleanup enqueue) is unchanged once both guards pass.

manuzzi:
## Summary

- Add "Delete permanently" to the dashboard gallery card menu, wiring up the
  existing (previously unused) `api.deleteGallery` client and the existing
  `DELETE /galleries/:id` route.
- Require the gallery to be archived first, enforced both in the menu and
  server-side (409 `gallery_not_archived`) — the API is also used outside this
  UI, so the guard lives in the API itself.
- Add a server-side guard against deleting a gallery with any print orders
  (409 `gallery_has_print_orders`) — `PrintOrderItem` rows reference `File` via
  a `Restrict` FK, so this also prevents an unfriendly DB constraint error.
- Add role-based delete authorization (`canDeleteGallery` in
  `gallery-access.ts`, deliberately stricter than the general access model):
  Admin → any visible gallery, Owner → only their own, Member → never. Exposed
  to the frontend as a `canDelete` field on each gallery, so the UI never
  computes permission logic itself.
- Confirmation requires typing the gallery's exact title via the shared
  `usePrompt()` primitive.
- New `apiError.*` i18n keys for the two new error cases.

Closes #24.

## Test plan

- [x] `npm run type-check` and `npm run check:i18n` in `apps/frontend`
- [x] `npm run test` in `apps/api` (9 new tests covering the permission +
      guard-order logic, 200/200 total passing)
- [x] Manual: archive + delete a gallery with no print orders as its owner,
      confirm it disappears and an audit log entry / cleanup job appear
- [x] Manual: wrong typed title → no request sent, inline error shown
- [ ] Manual: as owner, confirm the option is absent on a gallery you don't own
- [ ] Manual: as member, confirm the option never appears
- [ ] Manual: as admin, confirm delete works on any visible gallery
- [ ] Manual: direct API call on a live (non-archived) gallery → 409
- [ ] Manual: archived gallery with a print order → 409, gallery not deleted

Note: `npm run lint` (`next lint`) currently fails on this repo's Next.js 16
regardless of branch — pre-existing tooling issue, not introduced here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ Changes werden trotzdem klar als solche markiert. Details: `docs/VERSIONING.md`.
 
 ## [Unreleased]
 
+### Added
+
+- Galleries can now be permanently deleted. A gallery must be archived first, deletion respects studio roles (admins can delete any gallery they can see, owners only their own, members not at all), and any gallery with print orders is protected so order records are never lost.
+
 ## [0.74.1] - 2026-08-27
 
 A pull is enough. Only the main server is affected.
--- a/apps/api/src/lib/gallery-access.ts
+++ b/apps/api/src/lib/gallery-access.ts
@@ -8,8 +8,11 @@
  *   3. Er ist Studio-Inhaber (Rolle "owner") — Sicherheitsnetz, damit
  *      keine Galerien verwaisen, wenn ein Mitarbeiter das Studio verlässt
  *
- * "Zugriff" bedeutet volle Rechte: sehen, bearbeiten, löschen, freigeben.
- * Wer keinen Zugriff hat, sieht die Galerie gar nicht (404).
+ * "Zugriff" bedeutet volle Rechte: sehen, bearbeiten, freigeben. Wer keinen
+ * Zugriff hat, sieht die Galerie gar nicht (404).
+ *
+ * Ausnahme: das endgueltige Loeschen einer Galerie folgt einer eigenen,
+ * strengeren Regel — siehe canDeleteGallery() weiter unten.
  *
  * WICHTIG: Diese Helper sind die EINZIGE Stelle, an der das Modell
  * definiert ist. Alle Routen, die auf Galerien (oder deren Dateien,
@@ -55,6 +58,8 @@ export const galleryRelationAccessWhere = galleryAccessWhere;
  * Erwartet `ownerId` und — sofern relevant — die `collaborators` als
  * `{ userId }[]`. Fehlt `collaborators`, zählt nur Ersteller + Owner-Rolle
  * (der Aufrufer muss die Relation dann mitladen, wenn Freigaben zählen).
+ *
+ * Deckt NICHT das Loeschen ab — dafuer gilt canDeleteGallery().
  */
 export function canAccessGallery(
   s: SessionContext,
@@ -71,3 +76,25 @@ export function canAccessGallery(
  * das auch — die Prüfung ist daher identisch zu canAccessGallery.
  */
 export const canManageGalleryCollaborators = canAccessGallery;
+
+/**
+ * Delete-Berechtigung — bewusst STRENGER als canAccessGallery, das sonst
+ * fuer alle Aktionen gilt (siehe Modul-Kommentar oben). Produktentscheidung:
+ *
+ *   - Admin darf jede Galerie loeschen, die er ohnehin sehen darf
+ *     (galleryAccessWhere bleibt unveraendert — keine Ausweitung der
+ *     Sichtbarkeit, nur der Loesch-Berechtigung innerhalb dessen).
+ *   - Owner darf nur eigene Galerien loeschen (ownerId === eigene id) —
+ *     auch eine Collaborator-Freigabe auf eine fremde Galerie reicht
+ *     dafuer NICHT, obwohl sie fuer canAccessGallery genuegen wuerde.
+ *   - Member darf nie loeschen, unabhaengig von Ersteller- oder
+ *     Freigabe-Status.
+ */
+export function canDeleteGallery(
+  s: SessionContext,
+  gallery: { ownerId: string }
+): boolean {
+  if (s.user.role === "admin") return true;
+  if (s.user.role === "owner") return gallery.ownerId === s.user.id;
+  return false;
+}
--- a/apps/api/src/routes/collections.ts
+++ b/apps/api/src/routes/collections.ts
@@ -18,6 +18,7 @@ import {
   smartCollectionFilterSchema,
   buildWhereClause,
 } from "../services/smart-collection-filter.js";
+import { canDeleteGallery } from "../lib/gallery-access.js";
 
 const createSchema = z.object({
   name: z.string().min(1).max(120),
@@ -201,6 +202,7 @@ export async function registerCollectionRoutes(app: FastifyInstance) {
           watermarkEnabled: true,
           createdAt: true,
           updatedAt: true,
+          ownerId: true,
           _count: { select: { files: true } },
           tags: {
             select: {
@@ -223,6 +225,7 @@ export async function registerCollectionRoutes(app: FastifyInstance) {
           createdAt: g.createdAt,
           updatedAt: g.updatedAt,
           fileCount: g._count.files,
+          canDelete: canDeleteGallery(s, g),
           tags: g.tags.map((gt) => gt.tag),
         })),
       };
--- a/apps/api/src/routes/galleries.delete.test.ts
+++ b/apps/api/src/routes/galleries.delete.test.ts
@@ -1,0 +1,93 @@
+/**
+ * Tests für die Delete-Guards von DELETE /galleries/:id.
+ *
+ * canDeleteGallery() wird direkt aus gallery-access.ts importiert — die
+ * Datei hat nur Typ-Imports (Prisma, SessionContext), also keine
+ * Laufzeit-Abhängigkeit auf db.ts/config.ts (die beim Import fail-fast
+ * echte Env-Vars/eine DB-Verbindung brauchen). Die Guard-Reihenfolge
+ * (Berechtigung → archiviert → Druckbestellungen) selbst lebt inline in
+ * der Route und wird hier als Kopie geprüft — Synchronisation per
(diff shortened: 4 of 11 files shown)


manuzzi:
## Review of the automated Copilot findings

I checked both issues Copilot raised against the actual source rather than taking them at face value:

**1. `apiError.deleteNotAllowed` missing (flagged 4×, once per locale file) — confirmed, real bug.**
`useErrorText()` (`error-i18n.ts`) only falls back to `err.message` when no `apiError.<code>` key matches. Our new `403 delete_not_allowed` response sends `{ error: "delete_not_allowed" }` with no `message` field, so `err.message` ends up being the raw code itself — if this path is ever hit (e.g. stale `canDelete` after a role/ownership change between page load and click), the user would see the literal string `delete_not_allowed` in a toast. My assumption while planning this ("falls back to the generic `common.error`") was incorrect — there is no such generic fallback for `ApiError` instances, only for non-`ApiError` values. Will add `apiError.deleteNotAllowed` to all 4 locale files in a follow-up commit.

**2. Case-insensitive title match — assessed as by design, not a bug, but flagging for a second opinion.**
`DangerZone.tsx` (studio-level deletion — the one existing precedent for typed-confirmation in this codebase) does the exact same `trim().toLowerCase()` comparison, with an explicit comment: *"Studio-Name-Echo ist case-insensitive (UX), Backend macht das nochmal defensiv."* This PR's gallery-delete confirmation mirrors that intentionally. The i18n copy itself never claims exact/case-sensitive matching — only my issue/PR description used the word "exact," which was imprecise wording on my part, not a spec. I'd lean toward leaving the comparison as-is for consistency with the existing pattern and just softening "exact" in the issue/PR text, but this is a legitimate style question rather than a clear defect — open to the maintainer's call.

No other issues found on an independent pass of the diff. Not pushing any changes yet — will apply the `apiError.deleteNotAllowed` fix once confirmed.

## Finnish translation review

@canja006 — since you authored the original Finnish translation (`feat(i18n): add Finnish UI translation`), could you sanity-check the new `fi.ts` strings this PR adds? Produced without native review, so grammar/phrasing corrections are very welcome:

- `studio.deletePermanently`: "Poista pysyvästi"
- `studio.deleteConfirmTitle`: "Poista galleria pysyvästi"
- `studio.deleteConfirmMessage`: "Tämä poistaa pysyvästi galleria ”{title}” ja kaiken sen sisällön — tiedostot, jakolinkit, kommentit ja arvostelut. Toimintoa ei voi perua.\n\nVahvista kirjoittamalla gallerian nimi."
- `studio.deleteTitleMismatch`: "Kirjoittamasi teksti ei vastaa gallerian nimeä. Mitään ei poistettu."
- `apiError.galleryHasPrintOrders`: "Tällä gallerialla on tulostustilauksia, eikä sitä voi poistaa."
- `apiError.galleryNotArchived`: "Arkistoi galleria ennen sen pysyvää poistamista."

manuzzi:
Pushed 9232c2d: adds `apiError.deleteNotAllowed` to all 4 locale files, fixing the raw-error-code leak Copilot flagged. `check:i18n` (2600/2600 keys across all dictionaries) and `type-check` pass, full backend suite still 200/200.

Left the case-insensitive title match as-is per the earlier discussion (matches `DangerZone.tsx`'s existing precedent) — happy to revisit if there's disagreement there.

canja006:
Checked all seven — six are natural Finnish as written, nice work. One grammar fix needed in `studio.deleteConfirmMessage`: "Tämä poistaa pysyvästi galleria ”{title}”" → "Tämä poistaa pysyvästi **gallerian** ”{title}”" (the object needs the genitive-accusative case). Everything else, including the terminology, matches the existing fi.ts conventions — "tulostustilauksia" is consistent with `notifEventPrintOrderDesc`.

manuzzi:
Thank @canja006, I'll update that

manuzzi:
Pushed 308af03 — fixed, thanks @canja006: "galleria" → "gallerian" in `studio.deleteConfirmMessage`. `check:i18n` still passes (2600/2600).

markusthiel:
Thanks for this — the guard work is exactly right, and I like that `canDelete` is computed server-side so the UI never carries permission logic. The two 409 guards I'd merge as they are.

The role matrix needs adjusting, and I think the naming is to blame rather than anything you did.

In Lumio `owner` is the *studio owner* — the top role, not a gallery-level one. `galleryAccessWhere()` returns an empty fragment for it, so an owner sees every gallery in the studio; `admin` and `member` are both scoped to their own galleries plus collaborations. The team UI says the same ("Owner — can additionally assign owner roles"). Plenty of products use `admin` as the higher of the two, so reading it that way is an easy call to make — it just doesn't match this codebase.

As written, the studio owner would be the one person unable to delete a gallery an employee created, while an admin could delete any gallery shared with them. That collides with why the owner role has gallery-wide reach at all — the comment above `galleryAccessWhere()` calls it a safety net so galleries don't end up orphaned when an employee leaves. Cleaning up after someone has left is the exact case, and the current rule locks the owner out of it.

For deletion I don't want a distinction between owner and admin: an admin is an admin. Which makes this simpler than the matrix I put in #24, because the visibility check already does the scoping. The lookup with `galleryAccessWhere` runs first and 404s, so `canDeleteGallery` only ever sees a gallery the caller can already reach. The rule is then just "not a member":

```ts
/**
 * Delete-Berechtigung. Owner und Admin duerfen loeschen, Member nie.
 * Die REICHWEITE ergibt sich aus galleryAccessWhere(), das in der Route
 * vorher greift und sonst 404 liefert — fuer den Owner also jede
 * Galerie des Studios, fuer den Admin seine eigenen plus Freigaben.
 * Deshalb braucht diese Funktion die Galerie selbst nicht.
 */
export function canDeleteGallery(s: SessionContext): boolean {
  return s.user.role === "owner" || s.user.role === "admin";
}
```

That drops the `gallery` parameter, so the call in the list route becomes `canDeleteGallery(s)`. Worth noting it ends up equivalent to the predicate `requireOwner()` already uses (owner or admin) — I'd still keep it as its own named helper with the comment, because the reason it holds is the 404-first ordering, and that's worth stating where someone will read it.

One consequence I want to name rather than leave implicit: an admin will be able to permanently delete a colleague's gallery that was shared with them. I'm fine with that — a collaboration invite inside one studio isn't an arm's-length relationship, and archived-only plus the typed title are already two brakes. But it's a real widening compared to your version, and you've been closer to this code than I have this week. If you think the stricter rule is the right one, make the case and I'll listen.

#24's acceptance criteria carry my original assumption, so those two bullets want updating too — otherwise the issue and the code will disagree once this lands. I'll fix the issue text.

On the case-insensitive title match: agreed, leave it. Matching `DangerZone.tsx` is worth more than a stricter comparison here, and the user-facing copy never promised exact case. Softening "exact" in the issue text is enough.

markusthiel:
Picking this back up — I think it just needs the role change and it's done.

You agreed over on #29 that the slug shouldn't be stricter than deletion, and implemented owner-or-admin there in `67ba664`. This PR is still on `308af030` from the 31st, so `canDeleteGallery` here is the version that has it the other way round.

Same shape as what you ended up with on #29: since `galleryAccessWhere()` already scopes what an admin can reach before this check runs, the rule is just "not a member", and the function doesn't need the gallery argument any more.

```ts
export function canDeleteGallery(s: SessionContext): boolean {
  return s.user.role === "owner" || s.user.role === "admin";
}
```

That plus the call site in the list route, and the two acceptance-criteria bullets in #24 — I said I'd fix those and haven't yet, so I'll do that rather than leave it on you.

Everything else here I'm happy with as it stands.

manuzzi:
Agreed on the shape — thanks for picking up the role-change and the two `#24` acceptance-criteria bullets yourself. Matches exactly what we landed on over in #29 (`galleryAccessWhere()` already scopes what an admin can reach, so the gallery argument on `canDeleteGallery` was redundant).

Nothing else from me on this one — ready to merge whenever that commit is in.

Issue #31, opened 2026-09-04
Lightroom plugin: model Lumio Gallery as a Collection Set with Chapters


manuzzi:
## Problem

The Lightroom Classic plugin (`apps/lightroom-plugin/lumio.lrdevplugin`) only supports a **flat** publish model today: one Published Collection = one Lumio Gallery, no grouping. This is hard-blocked in `LumioPublishService.lua` via `maxCollectionSetDepth = 0`, a deliberate fix for a past crash (`"attempt to call method 'hierarchyCreated' (a nil value)"`) that happened because the plugin had no Collection-Set-level callbacks at all.

Meanwhile, Lumio's backend already has a full "Chapter" concept server-side — modeled as `GallerySection` in the schema/API (`apps/api/prisma/schema.parisma` → `GallerySection`, `apps/api/src/routes/galleries.ts` Section-CRUD block) and shown as "Chapters" in the Studio UI. It has just never been wired up to the plugin.

## Proposal

Let a photographer create a **Lumio Gallery as a Lightroom Collection Set**:

1. Creating the Set auto-creates a default child Published Collection. Photos published there land in the gallery's existing unsectioned/default bucket — exactly like today's flat mode.
2. Every *additional* child collection added inside the Set becomes a real Chapter (`GallerySection`) on Lumio.
3. Today's flat (non-Set) publishing keeps working unchanged, relabeled **"Simple Gallery"** in the plugin UI to distinguish it from the new Set-based, multi-chapter **"Lumio Gallery"**.
4. Renaming/deleting a chapter-collection in Lightroom propagates to the corresponding `GallerySection` (rename → `PATCH` title, delete → `DELETE` section), to the extent the LrC SDK supports it (confirmed hooks: `renamePublishedCollection`, `deletePublishedCollection`).

## Non-goals

- **Gallery deletion stays Studio-only**, exactly as today (tracked separately). Deleting the Set in Lightroom must never delete the Gallery server-side — only child chapters clean up their own `GallerySection`.
- No backend/API changes are expected to be needed — `apps/api/src/routes/galleries.ts` already implements full Section CRUD, reachable via the same Bearer-token auth the plugin already uses for uploads.

## Implementation checklist

No CI/automated tests exist for this plugin — every step below is verified manually by loading the plugin into a real Lightroom Classic install.

- [ ] 1. Bump plugin version 0.3.0 → 0.4.0 (`Info.lua`)
[...]

Issue #30, opened 2026-09-02
Gallery slugs are globally unique across tenants


markusthiel:
`Gallery.slug` is `@unique` across the whole instance, and the public lookup in `routes/galleries.ts` is `findUnique({ where: { slug } })` with no tenant filter. Two consequences on a multi-tenant instance:

- Two studios cannot both use the same readable gallery name. Whoever claims `hochzeit` first holds it instance-wide, and the collision error reveals that some other studio is using it.
- A gallery belonging to studio B resolves under studio A's subdomain. Harmless while slugs are unguessable random strings, but it is not the behaviour we want to keep.

Neither matters on a single-tenant self-host, where every request falls through to the default tenant anyway. This is specifically about shared and SaaS instances.

### What needs to happen

1. `@@unique([tenantId, slug])` instead of `slug @unique`. The migration must `DROP INDEX` — Prisma's `@@unique` generates an index, not a constraint. No backfill: existing random slugs stay valid under the composite key.
2. Tenant-filter the public lookup (`findFirst` on `{ tenantId, slug }`).
3. **Prerequisite for 2:** make notification links tenant-aware. `services/notifier.ts` builds `${config.PUBLIC_URL}/g/<slug>`, and `PUBLIC_URL` is a single global env var. On a multi-tenant instance it points at a host that `resolveTenant()` cannot map to a tenant (the apex, or `studio.`, which is reserved), so a tenant-filtered lookup would break every link sent by email. Links need to be built from `customDomain ?? <tenant.slug>.<LUMIO_DOMAIN_BASE> ?? PUBLIC_URL`. The studio UI is already fine — it uses `window.location.origin`.

Worth auditing whether anything else builds a `/g/` link from `PUBLIC_URL` before this lands.

This blocks #28 / #29: custom slugs on a global namespace would create a land-grab that can't be undone once real client links depend on it.

manuzzi:
Went through this against the current code (not just the description) since it blocks #29, which I opened.

Confirms cleanly: `Gallery.slug` is `@unique` (`schema.prisma`), and the public lookups are bare `findUnique({ where: { slug } })` with no tenant filter — [`routes/galleries.ts`](../blob/main/apps/api/src/routes/galleries.ts) at the `GET /g/:slug`, `/unlock`, and `/assets/:kind` handlers.

Two things worth adding to the scope before this lands:

[...]

PR #32, opened 2026-09-06
feat(lightroom-plugin): model Lumio Gallery as a collection set with chapters

manuzzi:
## Summary

Implements #31: a published collection can now be a **Collection Set**, which becomes a **Chapters Gallery** on Lumio. Saving the Set immediately creates a "Default" child collection (photos published there land in the gallery's normal unsectioned bucket), and every additional child collection becomes a real Chapter (`GallerySection`) on Lumio, with its uploaded photos assigned to that chapter automatically. Renaming or deleting a chapter collection in Lightroom syncs to the corresponding chapter on Lumio, and "Show in Lumio" now also works on the Set itself.

A plain collection directly under the publish service keeps working exactly as before. No server-side changes were needed — this reuses the existing Studio Section API (`/galleries/:id/sections*`), which already accepts the same Bearer-token auth the plugin already uses for uploads.

Plugin version bumped to **0.4.2**.

**Diff scope note:** #27 has merged, and this branch has been rebased onto `main` (v0.82.0). The diff now contains only the Chapters Gallery work — the icon / anti-duplicate republish / hash-matching commits that used to show up here are already in `main`.

## What changed

- `Info.lua`: version 0.4.2
- `LumioApi.lua`: 4 new methods (`createSection`, `patchSection`, `deleteSection`, `assignFilesToSection`), thin wrappers around the existing Studio Section-CRUD routes
- `LumioPublishService.lua`: Collection Set support (`maxCollectionSetDepth = 1`), the three required set-level dialog callbacks, gallery inheritance from parent Set on first publish, chapter creation + batched/chunked file assignment, rename/delete sync for chapters, and a `goToPublishedCollection` fix that now handles both plain collections and Sets. Follow-ups from review: Set rename/delete is an explicit no-op (via a shared `isPublishedCollectionSet()` check on `:type()`, so a genuine `getCollectionInfoSummary()` failure on a real chapter is logged instead of taking the same silent path as a Set), the chapter's section id is recorded as its remote id, and a nil `fileAttributes()` result is guarded
- READMEs (DE/IT) and CHANGELOG updated. The CHANGELOG's `[Unreleased]` section carries a plug-in reinstall notice, since the plug-in folder loaded into Lightroom is not updated along with the server

## Known SDK limitations (not bugs, confirmed against the official Lightroom Classic SDK reference)

- The native "Create Published Collection..." menu entry, both at the top level and inside a Chapters Gallery, uses Lightroom's own generic wording rather than a custom name — this string is shared across both contexts with no way to vary it selectively, so a custom label would misname a Chapter as a "Simple Gallery" (or vice versa).
- The native "Create Smart Collection..." entry cannot be hidden — there is no SDK-level flag to remove it, only to rename it, so it's left with Lightroom's generic wording rather than a misleading Lumio-branded name.
- Deleting a chapter collection only syncs to Lumio when **Delete** (not **Leave on Service**) is chosen in Lightroom's own confirmation dialog. That is documented SDK behavior; the README says so.

## Test plan

No automated test coverage exists for this plugin (no CI job touches `apps/lightroom-plugin`, no Lua test framework) — everything below was verified manually against a real **Lightroom Classic 15.5.1** install on macOS, against a live Lumio instance, across several rounds of real-device testing and fixes:

- [x] Creating a new Collection Set (Chapters Gallery) shows the Lumio gallery-picker dialog and immediately creates a nested "Default" collection
- [x] Publishing through "Default" creates the Lumio gallery and lands photos in the unsectioned bucket
- [x] Adding an extra collection inside the Set and publishing creates a matching Chapter (`GallerySection`) and assigns its photos to it
- [x] Renaming a chapter collection syncs the chapter's title on Lumio
- [x] Deleting a chapter collection deletes the corresponding Lumio section (files fall back to unsectioned)
- [x] "Show in Lumio" works both on a chapter collection and on the Set itself
- [x] Existing flat/plain published collections continue to work unchanged
- [x] Plugin reloaded and re-tested after each fix round; several real bugs (wrong/non-existent SDK method names, an invalid same-write-transaction read, a redundant title field) were only caught this way and are already fixed in this PR
- [x] Independently tested by @canja006 on their own Lightroom Classic 15.5.1 / macOS setup, checked against the database and `Lumio.log` (see the comments), including the Set/chapter rename and delete matrix on `7d04d7c`. The later `type()` refactor (`241d07b`) and the rebase onto `main` have not been re-run on a second install
- [x] After the rebase onto `main`: the 18 rebased commits are patch-equivalent to the pre-rebase branch (`git range-diff`); only the CHANGELOG hunks differ, because #27's entries had moved to `[0.81.0]`. All plug-in Lua files compile under LuaJIT


manuzzi:
@copilot please review the latest changes.

canja006:
Thanks for the ping — happy to run this on my setup, the same way as on #27.

The earliest I can get to it is tomorrow. I'll go through the full list on Lightroom Classic 15.5.1 / macOS against my self-hosted instance: Set creation dialog, the right-click menus on the Set and on a chapter collection, the auto-created "Default" child, chapter create/rename/delete round-trips, and "Show in Lumio" on the Set itself. Report in the same format as before.

canja006:
Ran the Chapters Gallery flow on my install — Lightroom Classic 15.5.1 on macOS, self-hosted Lumio still at `052b0f9` (the api/worker diff of this branch against it is comment translations only, so the server side needed no rebuild), plug-in from `be39ca5` (0.4.1). Same 2774-photo catalog as on #27. Everything below is checked against the database (`gallery_sections`, `files.sectionId`) and `Lumio.log`, not just the panel.

**Menus and dialogs, since you asked specifically.** Right-click on the service shows `Create Published Collection…`, `Create Published Smart Collection…`, `Create Chapters Gallery…` — both SDK limitations exactly as described. The Set dialog shows the "Chapters Gallery" group box with the gallery picker. Inside the Set, right-click gives the generic `Create Published Collection…`, and its first-open dialog shows the Simple Gallery picker (the cosmetic gap you documented). Nothing surprising on 15.5.

**What works**

- `Create Chapters Gallery…` → Set created, `Default` child appears immediately, its settings are `{ isDefaultChapter = true }`.
- Publish 3 photos via `Default` → `POST /plugin/galleries` (new gallery), three uploads, all three with `sectionId = NULL`. The Set's own settings picked up `galleryId`/`gallerySlug`/`galleryMode`, so later chapters inherit the right gallery.
- New collection inside the Set, 3 photos → `POST /galleries/:id/sections` before the first upload, then a single batched `POST /sections/:id/files` → all three assigned to the chapter, no stray gallery created.
- Rename chapter → `PATCH /sections/:id` within the same second, title updated.
- "Show in Lumio" on the Set and on a chapter → Studio view of the gallery.
- Plain collections at the service root behave as before (re-publish into an existing flat collection: uploads + `PATCH` gallery).

## Chapter deletion never reaches Lumio in the case the README describes

README step 4 says: delete the collection → chapter deleted on Lumio, photos fall back to the unsectioned bucket. On a real install it depends entirely on which button the user presses in Lightroom's own confirmation dialog:

| Choice in LrC's delete dialog | `deletePublishedCollection` invoked? | On Lumio |
|---|---|---|
| **Leave on Service** | **No** — not a single line logged | Chapter stays, its photos stay assigned to it. Only the Lightroom collection disappears. |
| **Delete** | Yes | `DELETE /sections/:id` → chapter removed; the photos **remain** in the gallery, unsectioned (no `deletePhotosFromPublishedCollection`, no `photoIds` in `info`). |

This is documented SDK behaviour, not a 15.5 quirk — the reference for `deletePublishedCollection` says: *"this callback function is only invoked when the user clicks the 'Delete' button in the dialog which Lightroom presents to alert the user of what happens to photos which have been published via this published collection. If the user chooses to leave the photos in their published location, the function is not called."*

So the README outcome is reachable only through the button whose own dialog says the photos will be deleted from the service, while the choice a photographer will naturally make to keep their photos ("Leave on Service") leaves an orphaned chapter on Lumio. I verified with a log line as the first statement of `deletePublishedCollection`: silence for "Leave on Service"; for "Delete" the whole path runs (`info` keys `isDefaultCollection, name, parents, publishService, publishedCollection`, `getCollectionInfoSummary` fine, `sectionId` present, `DELETE` 200).

I don't see a way for the plug-in to cover the "Leave on Service" case — there is no hook. What it can do is say so. README/CHANGELOG should state that removing a chapter on Lumio requires choosing **Delete** in Lightroom's dialog, that the photos are nevertheless kept in the gallery (unsectioned), and that "Leave on Service" only removes the Lightroom collection — the chapter then has to be deleted in Studio.

## Deleting the Set

Deleting the whole Set (with "Delete") leaves the gallery untouched, as intended — but by accident. Lightroom invokes `deletePublishedCollection` for the Set as well, `info.publishedCollection` is then an `LrPublishedCollectionSet`, and the code calls `getCollectionInfoSummary()` on it:

```
attempt to call method 'getCollectionInfoSummary' (a nil value)
```

`LrTasks.pcall` swallows it, `collSettings` ends up nil, nothing happens. An explicit early return for Sets (or `getCollectionSetInfoSummary`, the way `getParentSetSettings` already does it) would make the non-goal hold on purpose rather than by exception.

## Two small things

- `remoteId` is never set — `remoteCollectionId` is NULL for every Lumio collection in the catalog. The current logic doesn't need it, but `publishedCollection:setRemoteId(sectionId)` inside the existing "assign chapter" write block works (tested) and gives `info.remoteId` in the rename/delete callbacks for free. Optional.
- One photo whose master was offline gave "2 succeeded, 1 failed … The file could not be found". That's Lightroom's render failing; the plug-in reported it cleanly.

Happy to re-run the delete cases on a follow-up commit.

manuzzi:
@canja006 thank you for the thorough follow-up testing — checking against the DB and log, not just the panel, caught things I wouldn't have found otherwise. Pushed two commits addressing your report:

- `e8d1451`: `renamePublishedCollection`/`deletePublishedCollection` now explicitly check for a Collection Set and return, instead of relying on `getCollectionInfoSummary()` throwing and the `pcall` swallowing it — same end result, but by design now, not by accident. Also added `setRemoteId(sectionId)` in the chapter-assignment write block, as you verified.
- `7d04d7c`: README (DE/IT) and CHANGELOG now explain that a chapter's Section is only deleted on Lumio when "Delete" (not "Leave on Service") is chosen in Lightroom's own confirmation dialog, and that "Leave on Service" leaves an orphaned chapter that needs manual cleanup in Studio — exactly the SDK behavior you quoted.

Really appreciate you running the delete-case matrix in such detail. Let me know if you spot anything else on a follow-up pass.

manuzzi:
Forgot to bump to Plugin Version 0.4.2 for easier identification in LrC

canja006:
Re-ran the delete matrix on `7d04d7c` — plug-in reinstalled from the branch, Lightroom fully restarted, same server (`052b0f9`, unchanged as before).

- Chapter created inside a fresh Set → `POST /sections`, batched assignment as before, and the catalog now carries `remoteCollectionId = <sectionId>` for the chapter — `setRemoteId` lands where it should.
- Rename the Set itself → nothing sent to Lumio, gallery title untouched, no dialog.
- Delete the chapter with **Delete** → `DELETE /sections/:id`, section gone, all photos still in the gallery, unsectioned.
- Delete the whole Set with **Delete** → nothing sent, gallery intact (6 files, 0 sections), no error.

So the three cases from my report now behave the same as before but on purpose, and the README wording matches what Lightroom actually does.

One small thing on `e8d1451`: the Set detection still keys off `getCollectionInfoSummary()` failing inside the `pcall`, so a genuine failure of that call on a real chapter would take the same silent no-op path as a Set. The SDK gives you the type directly — `LrPublishedCollectionSet:type()` returns `"LrPublishedCollectionSet"` (and `"LrPublishedCollection"` for a plain collection) — so

```lua
if info.publishedCollection:type() == "LrPublishedCollectionSet" then return end
```

before the `pcall` would make the Set no-op a real check and leave the `pcall` failure as what it is, an error worth a `log:warn`. Not blocking; the behaviour is right either way.

Nothing else outstanding from my side.

manuzzi:
@canja006 thanks for the re-test — good to have all three cases confirmed clean after a full reinstall/restart, and that `setRemoteId` lands correctly.

Applied your `type()` suggestion in `241d07b`: added a shared `isPublishedCollectionSet()` helper that checks `info.publishedCollection:type()` (confirmed against the SDK reference to return `"LrPublishedCollection"` / `"LrPublishedCollectionSet"`) before ever calling `getCollectionInfoSummary()`. A genuine failure on a real chapter now logs a warning instead of silently taking the same path as a Set. Same behavior for the cases that matter, just an honest check instead of an inferred one — appreciate you flagging it even though it wasn't blocking.

Nothing else pending on my end either. Thanks again for all the real-device rounds on this one.

markusthiel:
Read through this and the whole exchange above. Nothing to add on the code — you two closed this out without me, and the review round is the kind I'd want as the norm here: @canja006 checking `gallery_sections` and `files.sectionId` in the database and the log rather than the panel, which is what surfaced the delete-dialog case at all; and the `:type()` catch, where the old code would have sent a genuine `getCollectionInfoSummary()` failure down the same silent path as a Set. Documenting the "Leave on Service" orphan instead of pretending it away was the right call too — the SDK does what it does.

One structural note. GitHub shows +2279 across 21 files here, but #27 is an ancestor of this branch, so most of that isn't yours: `schema.prisma`, `routes/plugin.ts`, `apps/worker/*` and `JpegXmp.lua` are all #27's. The actual contribution is +878 in 6 files, all inside the plug-in — which incidentally confirms the "no backend/API changes expected" note in #31.

So this goes in after #27, which still has three open items from Copilot's re-review on the 8th. Merging this first would pull #27 in silently, unfixed.

Makes sense that @canja006 tested against a server at `052b0f9` on the grounds that the API diff was comment translations only — that was accurate, and it means the two were effectively verified together.

No retargeting needed from my side, unlike the print-shop set; I mostly want the ordering written down somewhere. Queue is: #27 → this one.

manuzzi:
@markusthiel #27 has got an update, if successful I will rebase this PR with main branch and push agin to update this PR as in my flow this was straight after #27

markusthiel:
Merged as v0.83.0. That closes the Lightroom series — #27 and this one.

The diff dropping from +2279 across 21 files to +875 across 6 once #27 was in is the clearest possible illustration of why order mattered. This is exactly the 878 lines I measured back when the stack was still hidden.

Re-checked what changed since I read this on the 11th: one commit, the reinstall notice. `isPublishedCollectionSet()` with the explicit `:type()` check is unchanged, and that's the piece I'd have looked at hardest — it's the difference between a no-op that happens on purpose and one that happens because an exception got swallowed.

What makes this PR stand out isn't the feature, it's the trail: three rounds of real-device testing by @canja006, each one turning up something a reading of the code wouldn't have. Creating a collection inside the same catalog write transaction that created it, a non-existent catalog method called on the parent lookup, the delete dialog only syncing on "Delete" and not on "Leave on Service" — none of those are visible without Lightroom in front of you. Documenting that last one as SDK behaviour rather than pretending it works is the right call.

313 tests, `tsc` clean, i18n at 2699 keys, smoke test green before the tag.

Next is #30 PR2, and #25 still needs the two-line role change whenever you pass by it.

PR #34, opened 2026-09-07
feat(print-shop): bulk catalog import with tiered (quantity-break) pricing

### LINKED ISSUE #33: feat(print-shop): bulk catalog import with tiered (quantity-break) pricing
## Problem
The Print-Shop product catalog (`PrintProduct` / `PrintProductVariant`) can currently only be built by hand, one product/variant at a time, from the Studio UI. A studio that wants to load a real-world print price list (sizes, materials, prices — potentially hundreds of items) has no way to bulk-load it.

## Proposed solution
- A downloadable, Lumio-native JSON import template (not tied to any specific print lab's own price-list format) that a studio fills in and uploads from the Studio UI.
- Import is upsert-by-SKU: re-uploading an updated file updates existing products/variants that match by SKU, and creates new ones for anything unmatched — safe to re-run as a price list evolves.
- Non-conforming rows (missing dimensions, unknown category, etc.) are skipped individually with a warnings/errors report, rather than blocking the whole import.

## Why tiered pricing is in scope too
Real-world print price lists (e.g. a reference Italian print-lab list used to design this, ~320 products / ~375 size/material variants) almost universally price by **quantity break** — e.g. "1-19 copies: €0.35/pc, 20-99: €0.30/pc, ... 400+: €0.16/pc" for the *same* product/size. Lumio's data model today only supports a single flat price per variant, so importing such a list faithfully isn't possible without adding real quantity-tiered pricing support:
- New `PrintProductVariantPriceTier` model (additive migration, no backfill).
- Checkout pricing engine (`priceCart`) resolves the correct per-unit price from the ordered quantity.
- Studio variant editor gets a flat-vs-tiered pricing mode.
- Customer-facing gallery ordering UI shows/uses the resolved per-unit price live as quantity changes.

## Scope (high level)
- Prisma schema + additive migration for price tiers.
- Shared tier-ladder validation, used identically by manual variant CRUD and the bulk importer.
- New import endpoints: template download, dry-run preview, commit (both share one validation/service function so preview == what gets committed).
- Studio UI: import flow (upload → preview/report → confirm) + tiered-pricing editor on variants.
- Customer-facing gallery ordering UI: tier-aware live pricing.
- i18n: EN/DE/IT/FI.

Implementing on `feat/print-catalog-import`.

manuzzi:
Closes #33

## Summary
- Studios can now bulk-create/update their print catalog from a downloadable JSON template instead of adding every product/variant by hand.
- Import is **upsert-by-SKU**: re-uploading an updated file updates matching products/variants instead of duplicating them; rows without a SKU always create new ones.
- Import is **best-effort**: a broken row (missing dimensions, invalid price, bad ladder, unknown category, duplicate SKU within the file) is skipped with an error in the report; the rest of the file still imports. Preview (`dryRun: true`) and commit share one analysis function, so what a studio previews is exactly what gets written.
- Added real **quantity-tiered pricing**, since real-world print price lists almost universally price by quantity break (e.g. 1–19 copies at one rate, down to 400+ at a much lower one) — the catalog previously only supported a single flat price per variant.
  - New `PrintProductVariantPriceTier` model, purely additive migration, no backfill.
  - Shared ladder validation/resolution (`services/print/pricing-tiers.ts`) used identically by manual variant editing and the importer, so the two paths can never disagree on what a valid ladder is.
  - Checkout pricing engine (`priceCart`) now resolves the correct per-unit price from the ordered quantity.
  - Studio variant editor gets a flat/tiered pricing mode with a tier-row editor.
  - Gallery ordering UI (product picker + cart) is tier-aware, live, before checkout; the server remains authoritative for the actual charge.
  - Customer-facing quantity cap raised from 99 to 999 so deep tiers are reachable.
- i18n: EN/DE/IT/FI, all new strings added in lockstep.

## Test plan
- [x] `vitest run` in `apps/api` — 229/229 tests pass, including 38 new unit tests covering tier-ladder validation, quantity resolution, and best-effort/upsert-by-SKU import logic.
- [x] `tsc --noEmit` clean in both `apps/api` and `apps/frontend`.
- [x] `npm run check:i18n` — all 4 locale dictionaries in sync, no dangling `t()` keys.
- [x] Applied the new migration against a fresh Postgres alongside all 78 prior migrations (`prisma migrate deploy`); confirmed **zero schema drift** against `schema.prisma` afterwards.
- [x] Booted the real API server against that database and exercised the actual HTTP endpoints end-to-end: downloaded the template, previewed and committed an import with a mix of valid/tiered/broken rows (confirmed best-effort skip + correct warnings), re-imported with a matching SKU (confirmed update-not-duplicate), and exercised the manual variant CRUD tiered-pricing edge cases (missing `priceCents` on flat-switch, invalid ladder, valid switch-back) — all behaved as designed, verified against the DB directly.
- [x] Called `priceCart()` directly against a real tiered variant across the full quantity range (1, 19, 20, 150, 400, 1000) — resolved prices matched the ladder exactly at every tier boundary.
- [x] Full browser walkthrough of the Studio import UI and the gallery checkout UI was not performed (would require a seeded auth session + S3/minio-backed file upload beyond this session's scope) — the underlying HTTP/DB behavior they call into was verified directly as above.


manuzzi:
Addressed all 4 review comments in 5f323b6:

- Import Zod schemas (`widthMm`/`heightMm`, tier `minQty`/`maxQty`/`unitPriceEur`) were re-validating ranges that `catalog-import.ts` already handles per-row — a single out-of-range value anywhere in the file threw a raw `ZodError`, which the global error handler turns into a **500 for the entire request** (verified: no `.validation`/`.statusCode` on a bare `ZodError`, so it doesn't hit the 400 branch). Loosened to type-only checks; range/ladder validation now exclusively lives in the per-row path.
- Dropped `products.min(1)` — an empty file now returns a clean zero-row report instead of failing the same way.
- `planVariant()` now validates negative `costEur` (`invalid_cost_eur`, row skipped), matching how `priceEur` was already handled and how the manual variant schema treats `costCents`.
- Studio tier-row editor now caps at 20 rows client-side (matching the server's `MAX_TIERS`) and disables "+ Tier" at the limit — also covers the suppressed comment.

Re-verified end-to-end against a real Postgres + live server: a request with a negative width, a negative tier price, and `minQty: 0` in the same file now returns 200 with the good row created and each bad row individually reported and skipped (previously 500'd the whole request). Full suite: 231/231 tests passing, `tsc` clean.

markusthiel:
Six PRs in two days is a lot of ground covered — thanks. Before reviewing the features themselves I went through how they fit together, because the stack isn't visible from the outside and I'd have merged them in the wrong order.

**#40, #42 and #44 all have this branch as an ancestor.** They carry its six commits, so GitHub is showing their combined size rather than their own:

| PR | shown | actually its own |
|---|---|---|
| #40 bulk select | +3160 / 24 files | +188 / 5 |
| #42 finish options | +3754 / 26 | +795 / 19 |
| #44 order exports | +3527 / 24 | +560 / 9 |

All six target `main` and report mergeable, which is the trap: merging #42 first would pull the catalog import in silently and leave this PR a no-op.

Could you retarget #40, #42 and #44 onto this branch? Then the diffs show what each one actually adds, and the dependency is visible instead of inferred.

**Three conflicts that no merge order avoids.** `apps/api/src/services/print/orders.test.ts` doesn't exist on `main`, and #34, #36 and #38 each create it from scratch — 113, 37 and 63 lines. That's an add/add conflict, which is why almost every pair collides there. Whoever lands first creates the file; the other two need to append their cases. Separately, #40 and #42 both touch `print-shop/page.tsx` and conflict there.

I'd rather you rebased those than have me resolve them at merge time — you know which test cases belong together, and I'd be guessing in a conflict editor.

**Migrations are clean**, for the record: four of them, ascending timestamps, each on its own concern, none touching the same table. `20260907100000_print_variant_price_tiers` shows up in three PRs but it's byte-identical — just this branch travelling along.

Proposed order once the retargeting is done: this one, then #40, #44, #42, #36, #38. Feature reviews to follow — I didn't want to comment on six PRs while the shape underneath them was still in question.

markusthiel:
Read this one properly rather than skimming, since it's the base of the stack and it touches money. It holds up well.

What I checked specifically, because these are the places pricing code usually goes wrong:

- **The server never trusts the client.** `priceCart()` loads variants and their tiers from the DB and recomputes; `lib/print-pricing.ts` is preview-only and says so at the top. Right split, and worth the duplication.
- **`resolveCartItemPricing` is pure and DB-free**, with the Prisma I/O left in `priceCart()`. That's why the tier math is directly testable, and it shows in the test count.
- **Per-format aggregation is correct and non-obvious** — 15 photos each as one 10×15 is 15 units of that variant, not 15 lines that each see quantity 1. The comment explaining it is doing real work.
- **`eurToCents` flags precision loss instead of swallowing it**, and the warning surfaces at all three levels (tier, price, cost).
- **The ladder is validated as a unit**, not tier by tier, and a price that rises with quantity is a warning rather than an error. Both are the right calls — a step-up is almost always a typo, but it still resolves coherently, so blocking would be wrong.
- **Per-product transactions**, documented as deliberate. Good: a 600-row list failing halfway leaves whole products, not half a product.

Ran the gates against the branch here: 244 tests pass (28 files), `check:i18n` clean at 2635 keys in all four dictionaries, `tsc --noEmit` clean in both apps.

Two things, neither blocking.

**Is the cost side tiered in the source list?** `priceTiers` carries `unitPriceEur`, but `costEur` is a single flat value per variant. If a lab gives quantity breaks on what it charges *us* as well, then a 400-print order shows a deep customer price against a flat unit cost, and the margin the studio sees is wrong. You designed this against a real reference list — does it have volume breaks on cost, or is cost genuinely flat?

**Removing a photo silently raises the price of the others.** Falls out of the per-format tier, and the cart handles it correctly — `quantityByVariant` recomputes every line. But a customer who deletes one photo and watches five other lines get more expensive has no way to know why. There's a tier hint on the picker; is there anything on the cart lines? A short note when a line is priced above its single-unit price would cover it.

manuzzi:
Thanks @markusthiel for the review and all the comments.

About the cost tiers: good catch, the logic is based as one of my Suppliers (PhotoSì) has price tiers based on ordered quantity, [here am example](https://www.photosi.com/it/stampe/stampe-foto-digitali/stampe-classiche), and costs follow that logic too (e.g. cost are roughly 40% discounted from the final product price)

About the "hidden" price change, good catch, I'll look forward to introduce an alert in case that change will cause a tier downgrade

manuzzi:
Pushed 7c86f46 addressing both non-blocking points from the review:

**Cost per quantity-break tier.** `unitCostCents` now rides along the existing price tier row (same `minQty`/`maxQty`) instead of a separate cost ladder — matches how the PhotoSì-style reference list actually shapes a row (one breakpoint, both what we charge and what it costs us). All-or-nothing per ladder (a partial set is rejected as a row/request error, same treatment as the rest of the ladder validation); Studio-only — never priced into checkout, and confirmed not exposed on the public gallery catalog endpoint, which already `select`s only `minQty/maxQty/unitPriceCents`. Wired through the manual variant editor (new "cost also varies by quantity" toggle, off by default so existing tiered variants are unaffected), the bulk importer (`priceTiers[].costEur`, documented in the downloadable template with a worked example), and the product list badge.

**Silent tier downgrade on removal.** The cart now shows a dismissible notice when removing an item — or lowering its quantity — pushes the remaining lines of that format into a worse tier, since that raises their price with no visible cause otherwise. Non-blocking, clears itself if the next cart change doesn't also cause a downgrade.

Verified: 253/253 API tests passing (added coverage for the new cost-tier paths in `pricing-tiers.test.ts` / `catalog-import.test.ts`), `tsc --noEmit` clean in both apps, `check:i18n` clean at 2641 keys across all four dictionaries.

Not verified here: no local Postgres in this environment, so the new migration hasn't been applied/exercised against a live DB, and there's no frontend test runner configured in this repo yet, so `willDowngradeTier` only has API-adjacent coverage by analogy, not its own unit test.

manuzzi:
Frontend now tested on my personal front-end, downgrade alert fully working

PR #36, opened 2026-09-08
feat(print-shop): pending-payment step and payment reference for offline invoice orders

### LINKED ISSUE #35: feat(print-shop): pending-payment step and payment reference for offline invoice orders
## Problem

`createOrder()` (`apps/api/src/services/print/orders.ts`) sets an order's status straight to `paid` at creation time whenever the checkout used `paymentMode: "offline_invoice"` (self-print / no Stripe Connect configured). There is no manual confirmation step, and no way to record a receipt/invoice reference against the payment.

For a studio that self-prints and invoices offline, every order is immediately "paid" the moment a guest submits the cart, before any money has actually changed hands.

## Proposed solution

- `createOrder()` starts **every** order (both `paymentMode` values) at `pending_payment`, not just `stripe_connect` ones. `stripe_connect` orders are unaffected — the change only removes the offline_invoice fast-path to `paid`.
- Add a `paymentReference` column to `PrintOrder`. The existing `mark_paid` transition (`transitionOrder()` in `orders.ts`, exposed via `POST /print-shop/orders/:id/transitions`) requires this field when confirming payment on an `offline_invoice` order — Stripe orders keep using `stripeChargeId` as their natural reference, no duplicate manual input.
- Studio order-detail UI: the "mark paid" action becomes a small inline prompt (reference field) instead of a one-click button, for offline_invoice orders only.
- The reference is shown in the order detail and included in the CSV export already added for invoicing.

No changes to the `stripe_connect` flow, no new order-status values (reuses the existing `pending_payment`/`paid` states and `transitionOrder` state machine).

manuzzi:
## Summary

- `offline_invoice` checkouts (self-print, no Stripe Connect) previously landed straight on `paid` the instant a guest submitted the cart — no manual confirmation, no way to record a receipt/invoice reference. They now start `pending_payment`, exactly like `stripe_connect` orders.
- Confirming payment on an `offline_invoice` order now requires a `paymentReference` (added as a nullable column on `PrintOrder`), passed on the existing `mark_paid` transition (`POST /print-shop/orders/:id/transitions`). `stripe_connect` orders are unaffected — they keep using `stripeChargeId` as their natural reference, no new input required.
- Studio order-detail page: the "Mark as paid" action opens an inline prompt for the reference when the order is `offline_invoice`; the stored reference is shown next to the payment-mode label.
- Fixes a related latent bug: `transitionOrder()`'s `mark_paid` branch sent the "paid" confirmation mail directly but never set the `mails_sent_paid` idempotency marker that the `print-mail-sweeper` relies on, so the same mail would be re-sent ~30s later by the sweeper. This mattered little while `offline_invoice` bypassed `mark_paid` entirely, but now every `offline_invoice` order goes through this exact path, so it's fixed as part of this change (marker is now set before the mail fires, mirroring the pattern the old inline `createOrder()` code already used).

No new order-status values — reuses the existing `pending_payment` → `paid` states and `transitionOrder` state machine (`apps/api/src/services/print/orders.ts`).

Closes #35

## Test plan

- [x] `npx vitest run` — 195/195 tests pass (25 files), including 4 new tests for the pure `isMissingRequiredPaymentReference` helper
- [x] Live-verified against a throwaway Postgres instance by calling `createOrder`/`transitionOrder` directly:
  - `offline_invoice` order created → `pending_payment`, `paidAt: null`
  - `mark_paid` without `paymentReference` on that order → rejected, status unchanged
  - `mark_paid` with `paymentReference: "INV-2026-042"` → `paid`, `paidAt` set, reference stored
  - exactly one `mails_sent_paid` event recorded (confirms the double-mail fix — checked the actual mail log, one guest + one studio mail, not two)
  - `stripe_connect` order → still starts `pending_payment`; `mark_paid` still succeeds with no reference required


markusthiel:
Now that #34 is on `main`, this only conflicts on `orders.test.ts` — the file exists there with the catalog-import cases, so append yours rather than recreating it. Review follows once it's rebased. Details on #42.

manuzzi:
Rebased onto current `main` (`6b230f8`, force-pushed).

**The conflict:** `orders.test.ts` only, add/add — the file didn't exist yet when this branch forked, so both this branch (`isMissingRequiredPaymentReference` tests) and `main` via #34 (`resolveCartItemPricing` tests) created it independently. Appended mine after the existing describe block, merged the two import lines, no logic overlap.

**The last open Copilot finding:** the `mark_paid` validation error (`"paymentReference is required to mark an offline_invoice order as paid"`) was the only English string in `transitionOrder()` — every other error in this service is German (`Order nicht gefunden`, `Cart leer`, etc.) and this one is returned to the client as-is. Translated it to match.

**Verified after rebase:** `tsc --noEmit` clean in `apps/api` and `apps/frontend`, `vitest run` — 280/280 in `apps/api`, `npm run check:i18n` clean. Didn't re-run the live-Postgres checks from the original test plan (no DB in this environment) — those covered the payment-reference/mail-idempotency behavior that the rebase didn't touch.

Ready for review.

PR #38, opened 2026-09-08
feat(print-shop): in-store pickup delivery option and forked fulfillment checklist

### LINKED ISSUE #37: feat(print-shop): in-store pickup delivery option and forked fulfillment checklist
## Problem

1. Delivery is courier-only today (`ShippingMethod` has no way to mark itself as a pickup option), so a studio that also offers in-store pickup has no way to model it — every checkout collects a full shipping address even when the photographer just hands the prints over in person.
2. The order fulfillment lifecycle (`pending_payment → paid → in_production → shipped → delivered`) has no pickup-aware step, and the Studio order-detail page shows only the single next available action, not an at-a-glance view of overall progress.

## Proposed solution

- `ShippingMethod.isPickup` (new boolean): when set, checkout skips collecting a shipping address for that method.
- `PrintOrder.isPickupDelivery` (new boolean, snapshotted from the chosen method at order-creation time — same audit-snapshot reasoning as the existing `shippingAddress` JSON blob) drives a fork in the fulfillment state machine: after `in_production`, courier orders move through `shipped → delivered` as today, pickup orders move through a new `ready_for_pickup → delivered` path instead (shared terminal `delivered` status, relabeled "picked up" for pickup orders — no need for a whole separate terminal value).
- `transitionOrder()`'s allowed-transitions map becomes a pure, unit-tested `allowedTransitionsFor(status, isPickupDelivery)` function so a courier order can never be marked `mark_ready_for_pickup` and vice versa.
- Studio order-detail page: a new read-only-but-clickable "Fulfillment" checklist (4 steps: Ordered → Printing → Shipped/Ready for pickup → Delivered/Picked up) that mirrors the existing setup-checklist visual pattern already used elsewhere in the print-shop Studio UI. Clicking the current step fires the exact same transition as the equivalent action button — the checklist is a view onto the real state machine, not an independent toggle, so it can't drift from what actually happened (kept in one shared `runTransition()` handler with the action-button row).
- Checkout UI hides the shipping-address fields (keeping name/email/phone) when the selected method is a pickup method, showing a short note instead.

No new terminal states beyond the existing 8; purely additive schema changes (2 new nullable/defaulted columns, 1 existing column widened from required to nullable).

## Test plan

- [x] `npx vitest run` — 197/197 tests pass, including 6 new tests for `allowedTransitionsFor` covering both paths and the shared terminal status
- [x] Live-verified against a throwaway Postgres instance:
  - a courier checkout without a shipping address is rejected; a pickup checkout without one succeeds
  - `priceCart()`'s `isPickupDelivery` flag correctly reflects the chosen shipping method (and defaults false with no method selected, for the price-preview-before-choosing-shipping case)
  - a pickup order's `mark_shipped` is rejected at `in_production`; `mark_ready_for_pickup` succeeds, sets `readyForPickupAt`, and the order still reaches the shared `delivered` status afterward
  - a courier order's `mark_ready_for_pickup` is rejected at `in_production`; `mark_shipped` → `mark_delivered` works as before

manuzzi:
## Summary

- `ShippingMethod.isPickup` (new boolean): studios can now mark a shipping method as in-store pickup. Checkout skips collecting a shipping address for that method (`PrintOrder.shippingAddress` is now nullable — a widening, additive change).
- `PrintOrder.isPickupDelivery` (new boolean) snapshots the chosen method's `isPickup` flag at order-creation time — same audit-snapshot reasoning the existing `shippingAddress` JSON blob already uses, so historical orders stay stable even if a shipping method's config changes later.
- That flag forks the fulfillment state machine after `in_production`: courier orders keep going through `shipped → delivered`; pickup orders go through a new `ready_for_pickup → delivered` path instead. Both converge on the same terminal `delivered` status (relabeled "picked up" for pickup orders in the UI) — no need for a whole separate terminal value. The allowed-transitions logic is now the pure, exported, unit-tested `allowedTransitionsFor(status, isPickupDelivery)` in `apps/api/src/services/print/orders.ts`, so a courier order can never be marked `mark_ready_for_pickup` and a pickup order can never be marked `mark_shipped`.
- Studio order-detail page gets a new "Fulfillment" checklist section (Ordered / Printing / Shipped-or-Ready-for-pickup / Delivered-or-Picked-up), visually mirroring the existing print-shop setup checklist. It's a read-at-a-glance view onto the real state machine, not an independent toggle: clicking the current step fires the exact same transition as the equivalent action button, through one shared `runTransition()` handler, so the checklist can never drift from what the order actually did. Cancel/refund stay in a separate action row since they're side-exits, not checklist steps.
- Checkout UI: when the selected shipping method is a pickup method, the address fields (street/postal/city/country) are hidden in favor of a short note; name/email/phone are still collected.

Closes #37

## Test plan

- [x] `npx vitest run` — 197/197 tests pass, including 6 new tests for `allowedTransitionsFor`
- [x] Live-verified against a throwaway Postgres instance (`createOrder`/`transitionOrder`/`priceCart` called directly):
  - courier checkout without a shipping address → rejected; pickup checkout without one → succeeds
  - `priceCart()`'s `isPickupDelivery` correctly reflects the chosen method (and defaults `false` with no method chosen yet, for the price-preview case)
  - pickup order: `mark_shipped` rejected at `in_production`; `mark_ready_for_pickup` succeeds, sets `readyForPickupAt`, order still reaches shared `delivered`
  - courier order: `mark_ready_for_pickup` rejected at `in_production`; `mark_shipped → mark_delivered` unaffected


markusthiel:
Now that #34 is on `main`, this only conflicts on `orders.test.ts` — the file exists there with the catalog-import cases, so append yours rather than recreating it. Review follows once it's rebased. Details on #42.

manuzzi:
Rebased onto current `main` (`226247e`, force-pushed).

**The conflict:** `orders.test.ts` only, same add/add pattern as #36 — this branch's `allowedTransitionsFor` tests and `main`'s (via #34) `resolveCartItemPricing` tests both created the file from scratch with no shared ground. Merged imports, appended the describe block.

**The last open Copilot finding:** the fulfillment checklist built its step ranking from a plain array that only ever contained *one* of `shipped`/`ready_for_pickup`, picked by `order.isPickupDelivery`. If that flag and `order.status` ever disagreed, `rank.indexOf(order.status)` returned `-1` and the whole checklist silently disappeared. `allowedTransitionsFor()` already treats that exact mismatch as tolerable by design (both statuses allow the same next transitions, per the "stale flag" test case) — the checklist should degrade the same way, not vanish. Replaced the array with a `STATUS_RANK` map that gives `shipped` and `ready_for_pickup` the same rank regardless of the flag.

(The other Copilot finding on `api.ts` — the `typeof input.shippingAddress` self-reference flagged as "will fail type-checking" — is a false positive: verified with a real `tsc --noEmit` run, exit 0. No change needed there.)

**Verified after rebase:** `tsc --noEmit` clean in `apps/api` and `apps/frontend`, `vitest run` — 280/280 in `apps/api`, `npm run check:i18n` clean. Didn't re-run the live-Postgres checks from the original test plan (no DB in this environment) — those covered pickup/courier checkout and transition behavior that the rebase didn't touch.

Ready for review.

markusthiel:
Merged in v0.80.0. That closes the print-shop series — #34, #36, #38, #40, #42, #44 and #52 are all in.

What I checked here specifically, since this one widens a NOT NULL:

- `shippingAddress` becoming nullable is safe for existing rows, and the readers hold up: `formatAddress()` already returned an empty string for a non-object, and the export's pickup branch from #44 was waiting for exactly this case.
- The address requirement moved out of the Zod schema and into `createOrder()`, after the shipping method is loaded. That's the right place — whether an address is required depends on `isPickup`, which the request body can't be trusted for and the schema can't see.
- `allowedTransitionsFor(status, isPickupDelivery)` forks at `in_production` and both paths converge on `delivered`. Taking the flag from the order's own snapshot rather than looking the method up live means editing a shipping method later can't strand an in-flight order in a state it can no longer leave.

307 tests, `tsc` both apps, i18n 2699 keys, build.

Next is #30, per the sequence — and #49 is still waiting on your read of the four design questions, whenever you get to it.

PR #40, opened 2026-09-08
feat(print-shop): select multiple photos at once and add them to the cart in bulk

### LINKED ISSUE #39: feat(print-shop): select multiple photos at once and add them to the cart in bulk
## Problem

The gallery print-shop picker handles one photo at a time: browse → click a photo → choose product/variant/quantity → add to cart, repeated once per photo. Ordering "all the photos in 10x15" means repeating that whole flow for every single shot.

## Proposed solution

- Add a "select multiple" mode to the photo grid (`BrowseStep` in `apps/frontend/src/app/g/[slug]/print-shop/page.tsx`): tap to toggle selection, a floating "N selected · Add to cart" bar opens the picker scoped to the whole selection at once.
- `PickerDialog` generalizes from a single `file` prop to `files: PublicFile[]` (a single click still passes an array of one, so existing behavior is unchanged). Product/variant/quantity stay a single choice for the whole batch — the quantity field becomes "quantity per photo" — and confirming creates one cart line per selected photo in one shot.
- Crop stays single-photo-only in this pass: batch mode uses each photo's default framing, with a short note explaining that individual crop is only available when adding one photo at a time. A crop editor for N differently-shaped photos in one dialog wouldn't be usably designed for a first pass.
- The "at quantity X: price/pc" tier preview and the resulting subtotal account for `quantity × number of selected photos`, so the price shown matches what the server will actually charge once the quantity-break tiers resolve.

No backend/schema changes — this is purely a frontend UX change reusing the existing single-photo `POST /g/:slug/print-shop/checkout` cart-line shape (N lines instead of 1 per confirm).

manuzzi:
## Summary

- `BrowseStep` (the gallery photo grid in `apps/frontend/src/app/g/[slug]/print-shop/page.tsx`) gains a "select multiple" mode: tap photos to select them, then a floating "N selected · Add to cart" bar opens the picker scoped to the whole batch at once — instead of repeating the picker flow once per photo.
- `PickerDialog` generalizes its `file` prop to `files: PublicFile[]` (a normal single click still passes an array of one, so the existing one-photo flow is byte-for-byte unchanged). Product/variant/quantity stay one shared choice for the batch; the quantity field becomes "quantity per photo"; confirming creates one cart line per selected photo in a single action.
- Crop stays single-photo-only in this pass — batch mode uses each photo's default framing with a note explaining that individual crop needs adding photos one at a time (a crop editor across N differently-shaped photos in one dialog isn't a sensible v1 design).
- The tier-price preview ("at quantity X: price/pc") and the subtotal shown in the dialog now factor in `quantity × number of selected photos`, matching how the server's quantity-break tiers actually resolve at checkout.

Purely a frontend change — no schema/API changes. This branch is based on `feat/print-catalog-import` (PR #34, not yet merged) because it reuses `aggregateQuantityForVariant`/`unitPriceForQuantity` from `apps/frontend/src/lib/print-pricing.ts`, added there. The diff below therefore includes #34's changes until that merges; please review/merge #34 first, or diff against `feat/print-catalog-import` instead of `main` if reviewing standalone.

Closes #39

## Test plan

- [x] `npx tsc --noEmit` clean
- No frontend test convention exists in this repo for UI components in this area (confirmed against existing test files), and I don't have interactive browser access in this environment — this change was verified via type-checking and careful manual code review (tracing the bulk quantity/price math against the existing, already-tested `aggregateQuantityForVariant`/`unitPriceForQuantity`), not by clicking through it in a running browser. Flagging that explicitly rather than claiming UI verification I didn't do — worth a manual click-through before merging.


markusthiel:
Small and clean. A few things I specifically liked: `toggleSelectMode` clearing the selection so you can't leave stale ids behind, `files.filter(f => selectedIds.has(f.id))` degrading gracefully if the grid changes underneath, and `projectedQuantity` correctly becoming `quantity * files.length` so the tier preview reflects a bulk add. That last one is easy to forget and would have shown customers the wrong price.

Also noticed you picked up `shareFilename` for the thumbnails and cart names, so bulk orders respect the gallery's label mode. Thanks for reading that code rather than reaching for `f.filename`.

One thing I'd change, on the crop.

Disabling the crop editor for a multi-selection is right — a crop UI for N differently-framed photos in one dialog isn't usable. But the consequence is that bulk lines go out with `crop: null`, while the single-photo path always attaches a crop whenever the variant has a fixed `aspectRatio`. So the same variant produces order lines in two different shapes, and the studio is the one who has to sort it out: some lines say exactly how the customer wanted it framed, others say nothing at all. With `manual_self_print` being the working fulfilment path today, that's a person cropping by hand and guessing.

`defaultCropForAspect` already does the work and takes the photo's own dimensions, so it can just be called per file inside the `files.map()` instead of passing `null` — every line then carries a deterministic centered crop, and a customer who wants control over a particular photo orders that one on its own. Same UI, consistent data.

Worth a note in the dialog either way, I think: something saying photos will be cropped to fit the format automatically and can be adjusted by ordering individually. Prints are custom goods and don't come back.

PR #42, opened 2026-09-08
feat(print-shop): selectable finish options per variant (e.g. frame color)

### LINKED ISSUE #41: feat(print-shop): selectable finish options per variant (e.g. frame color)
## Problem

A `PrintProductVariant` has exactly one price/SKU. There's no way to offer the same physical format with alternative finishes — e.g. a poster with a black-frame or white-frame option — without duplicating the entire variant (dimensions, price, quantity-break tiers) once per finish, which also fragments the quantity-tier aggregation across what should count as the same format.

## Proposed solution

- New `PrintProductVariantFinishOption` model: `name`, optional `sku` (overrides the variant's own SKU on invoicing exports when selected), `priceDeltaCents` (added to the tier-resolved unit price). Additive/optional — a variant with none behaves exactly as today.
- Deliberately distinct from the existing `finishType` field, which is a studio-facing descriptive material label (matte/gloss/...) that customers never choose — finish options are a new, customer-selectable dimension.
- Quantity-break tiers still resolve per variant across every finish — a black-frame and white-frame order of the same poster still count toward the same quantity discount; the finish only adds its own flat surcharge on top, after the tier lookup.
- If a variant has any enabled finish options, selecting one is required before checkout (server-enforced in `resolveCartItemPricing`, not just a UI nicety) — no silent default that could skip a surcharge.
- Studio product editor gets a repeatable finish-option list (mirrors the existing tiered-pricing row editor). Customer picker gets a required finish selector when applicable, with the surcharge reflected in the live price preview.
- Bulk catalog import (`POST /print-shop/import/*`) supports an optional `finishOptions` array per variant, same best-effort-skip semantics as the rest of the importer.
- CSV order export gets a "Finish" column, and the SKU column falls back to the finish's own SKU first when one is set.

## Test plan

- [x] `npx vitest run` — 256/256 tests pass, including new coverage for the finish-surcharge pricing math and the importer's finish-option parsing/validation
- [x] Live-verified against a throwaway Postgres instance: surcharge resolves correctly, missing-selection is rejected, order-item snapshot (name/SKU) survives the finish option later being deleted (FK nulls, snapshot fields don't), and a bulk import creates finish options with correctly converted price deltas

manuzzi:
## Summary

- New `PrintProductVariantFinishOption` model (`apps/api/prisma/schema.prisma`): `name`, optional `sku` (overrides the variant's own SKU on invoicing exports when this finish is selected), `priceDeltaCents`. Additive/optional — a variant with none behaves exactly as today. Deliberately distinct from the existing `finishType` field, which is a studio-facing descriptive material label (matte/gloss/...) never chosen by the customer.
- Quantity-break tiers still resolve per variant across every finish — a black-frame and white-frame order of the same poster still count toward the same quantity discount. The finish only adds its own flat surcharge on top, after the tier lookup (`resolveCartItemPricing` in `apps/api/src/services/print/orders.ts`).
- Selecting a finish is required server-side whenever a variant has any enabled finish options — no silent default that could skip a surcharge, enforced in the same money-authoritative function checkout goes through.
- Studio product editor (`apps/frontend/src/app/studio/print-shop/products/page.tsx`) gets a repeatable finish-option row editor, mirroring the existing tiered-pricing editor pattern exactly (same add/remove/validate shape, capped at 20).
- Customer picker (`apps/frontend/src/app/g/[slug]/print-shop/page.tsx`) shows a required finish selector when the chosen variant has options, with the surcharge reflected in the live price preview and cart line.
- Bulk catalog import (`apps/api/src/services/print/catalog-import.ts`) supports an optional `finishOptions` array per variant — a broken entry skips the whole variant, same granularity as an invalid price-tier ladder. Template updated with a two-finish example.
- CSV order export gets a "Finish" column; the SKU column now prefers the selected finish's own SKU before falling back to the variant/product chain.

This branch is based on `feat/print-catalog-import` (PR #34, not yet merged) — it needs the tiered-pricing infrastructure added there (`PrintProductVariantPriceTier`, `resolveUnitPriceForQuantity`, the tier-row editor pattern it reuses). The diff below therefore includes #34's changes until that merges; please review/merge #34 first, or diff against `feat/print-catalog-import` instead of `main` if reviewing standalone.

Closes #41

## Test plan

- [x] `npx vitest run` — 256/256 tests pass, including new coverage for finish-surcharge pricing (aggregation, missing/invalid selection, zero-delta) and the importer's finish-option parsing (duplicate names, price-delta rounding)
- [x] Live-verified against a throwaway Postgres instance:
  - surcharge resolves correctly on top of the tiered unit price
  - ordering a variant with finish options without selecting one is rejected
  - an order's `finishOptionName`/`finishOptionSku` snapshot survives the finish option later being deleted (FK nulls via `onDelete: SetNull`, snapshot fields don't)
  - a bulk import creates finish options with correctly converted price deltas


markusthiel:
v0.76.0 is out with #34, #44, #52 and #40 merged, in that order. All four went in clean — and that makes the retargeting I asked for on #34 unnecessary. Once #34 was on `main`, the others just brought their own commits along. I should have tested the merge order before asking you to restructure branches; sorry for the detour.

Three things landed alongside that you should know about:

- The `zip_downloads` unique index from `20260520` was still there. `20260529` tried to drop it as a constraint, but it was created as an index, so `DROP CONSTRAINT IF EXISTS` silently did nothing. Your `source` migration was correct — the constraint *you* dropped was created with `ADD CONSTRAINT`, so that path works. I added a `DROP INDEX` alongside, otherwise the old index would have kept colliding on rows that differ only in `variant` or `source`.
- The red smoke tests on #52 and then on `main` were not your code. MinIO has closed anonymous pulls of `minio/minio` and `minio/mc` on Docker Hub — 401 even with a proper pull token. Switched both to `quay.io`; green since.
- #40 went in as-is. The `crop: null` point from my review stands, but it's a data-consistency nit, not a bug — I'll file it separately rather than hold the merge.

What's left is this one, and it needs a rebase onto current `main`. The conflicts are real rather than stale-base noise: your finish options and the tiered-cost commit both extend `catalog-import.ts`, the cart page overlaps with #40's bulk select, and the i18n files touch the same blocks. You know both sides of each of those, so I'd rather not resolve them for you.

#36 and #38 only collide on `orders.test.ts` — that file now exists on `main` with the catalog-import cases, so it's a matter of appending yours.

Reviews for all three follow once they're rebased; I've held off so far because reviewing code that's about to move around helps nobody.

manuzzi:
Rebased onto current `main` (`b6965f3`, force-pushed) and resolved the conflicts.

**The 7 conflicting files:**
- `catalog-import.ts`, `catalog-import.test.ts`, and all four `i18n/*.ts` files — all additive conflicts (both sides appended new content — `_readme` entries, `it(...)` test cases, translation keys — at the same insertion point). Kept both sides, no logic overlap.
- `g/[slug]/print-shop/page.tsx` (customer picker/cart) — the real one. #40's bulk-select turned `add()` into a `files.map(...)` and the price preview into a `projectedQuantity`/`quantityByVariant`-based calc; this branch added the finish-selection guard and surcharge on a still-single-file version of the same code. Merged them: bulk `add()` now carries `finishOptionId`/`finishOptionName` per item, and both the live preview and the cart-line price apply the finish surcharge on top of the bulk-aware tiered quantity, rather than one replacing the other.

**One more thing `tsc` caught, not git:** `GET /print-shop/orders/:id/export.md` was added on `main` by another already-merged PR while this branch was in flight, copy-pasting the `.csv` route's row-building as-is — so it never picked up `finishOptionName`/`finishOptionSku`. No conflict marker for it (the whole route arrived as "new" content from `main`'s side), but `OrderExportRow` now requires `finishName`, so it failed to compile. Fixed to mirror the `.csv` route exactly (same select, same SKU fallback order).

**Verified after rebase:** `tsc --noEmit` clean in `apps/api` and `apps/frontend`, `vitest run` — 288/288 in `apps/api`, `npm run check:i18n` clean. Didn't re-run the live-Postgres checks from the original test plan (no DB in this environment) — those covered pricing/import behavior that the rebase didn't touch.

Ready for review.

markusthiel:
Rebase is clean, reviewed the whole thing. Nothing blocking — this can go in.

What I checked, since it's price-affecting:

- **Client sends only the id; the server owns the delta.** `priceCart()` loads the variant's *enabled* options from the DB and looks the id up inside that list, so an id from another variant, or a disabled one, falls through. Right.
- **Required exactly when offered, no silent default.** Your comment names the reason — a surcharge could otherwise be skipped — and the reverse check (id given, variant has none) catches stale or forged ids. Both directions covered.
- **Delta on top of the tier-resolved price, not inside the tier lookup.** The schema comment explains why: ten black frames and five white ones are fifteen prints of that format. Agreed, and it's the kind of thing that would be wrong in a subtle way if done the other way.
- **Snapshot on the order line** with `SetNull` on the FK. So a renamed or deleted option can't change what a past order says.
- **Bulk path from #40** goes through the same `add()`, and `finishSelectionMissing` blocks it too. I specifically went looking for a way to bulk-add lines without a finish and get a server rejection; there isn't one.

Gates here: 288 tests, `tsc` both apps, i18n 2672 keys, build.

Two notes, neither needs a change now:

- Re-import recreates finish options with fresh ids, so every historical order line loses its FK link (the snapshot keeps name and SKU). You've documented that as a tradeoff, and I agree it is one — reporting by option would have to go via name rather than id. Worth remembering if someone later asks for "how many black frames did we sell".
- Same mechanism: a customer with an item in their cart during a re-import will get "finish option required" at checkout, because their id no longer exists. Rare, and the failure is the safe direction — rejects rather than misprices. Fine as is.

Merging.

PR #44, opened 2026-09-08
feat(print-shop): bulk photo ZIP download and Markdown order summary

### LINKED ISSUE #43: feat(print-shop): bulk photo ZIP download and Markdown order summary
## Problem

The Studio order-detail page can download one order photo at a time and a CSV of line items, but there's no way to get all of an order's photos in one download, and the CSV alone doesn't convey order-level context (customer, address, status) at a glance for handing to a lab or filing away.

## Proposed solution

- New `GET /print-shop/orders/:id/export.md` — a human-readable Markdown summary (customer, status, payment mode, shipping address/method, tracking, item table) built by a new pure `buildOrderSummaryMarkdown()` in `apps/api/src/services/print/order-export.ts`, reusing the same `OrderExportRow` shape as the existing CSV export.
- New `POST /print-shop/orders/:id/export-zip` — bundles every photo referenced by the order (deduplicated, since multiple lines can point at the same photo) into a ZIP, by calling the existing `requestZipDownload()` pipeline (`apps/api/src/services/zip.ts`) that already powers gallery-wide Studio ZIP downloads. No new download/build infrastructure — the resulting job is polled and downloaded through the existing generic `GET /galleries/:id/download/zip/:zipId` route, which looks a ZIP up by `(galleryId, zipId)` alone regardless of which endpoint created it.
- Studio order-detail page gets two new controls next to the existing CSV link: "Download summary" (MD) and "Download all photos" (ZIP, with the same request → poll → ready-link flow already used for gallery-wide Studio ZIP downloads).

## Test plan

- [x] `npx vitest run` — 253/253 tests pass, including 9 new tests for `buildOrderSummaryMarkdown` (heading, customer/status/payment fields, address formatting, pickup-vs-address branching, tracking conditionally shown, item table rendering, totals from the header not the rows, pipe-character escaping, missing-SKU rendering)
- [x] Live-verified against a throwaway Postgres + Redis instance: the exact query/mapping the `export.md` route uses produces correct Markdown for a real seeded order (verified customer, address, note, item rows, and totals all match); the `export-zip` route's dedup + `requestZipDownload()` call creates a correctly-scoped `ZipDownload` record and reuses the cached record on a second request for the same order

manuzzi:
## Summary

- `GET /print-shop/orders/:id/export.md` — human-readable order summary (customer, status, payment mode, shipping address/method, tracking, item table, totals), built by a new pure `buildOrderSummaryMarkdown()` in `apps/api/src/services/print/order-export.ts`. Reuses the existing `OrderExportRow` shape (same data as the CSV export) rather than introducing a parallel row type.
- `POST /print-shop/orders/:id/export-zip` — bundles every photo in the order (deduplicated — multiple lines can reference the same photo) into a ZIP by calling the existing `requestZipDownload()` pipeline in `apps/api/src/services/zip.ts`, the same one that already builds gallery-wide Studio ZIP downloads. No new build/download infrastructure: the resulting job is polled and downloaded through the existing generic `GET /galleries/:id/download/zip/:zipId` route, which looks a ZIP up by `(galleryId, zipId)` alone, regardless of which endpoint created it — confirmed by reading that route before adding this one, rather than assuming.
- Studio order-detail page: two new controls next to the existing CSV link — "Download summary" (direct link to the `.md`) and "Download all photos" (request → poll → ready-link, mirroring the exact UX pattern already used for gallery-wide Studio ZIP downloads in `GalleryFileTagFilter.tsx`).

Closes #43

This branch is based on `feat/print-catalog-import` (PR #34, not yet merged) — it needs the tiered-pricing/CSV-export infrastructure added there. The diff therefore includes #34's changes until that merges; please review/merge #34 first, or diff against `feat/print-catalog-import` instead of `main` if reviewing standalone. (Flagged by Copilot review — apologies, I omitted this disclosure here even though I added it on the other two PRs stacked the same way, #40 and #42.)

## Test plan

- [x] `npx vitest run` — 253/253 tests pass, including 9 new tests for `buildOrderSummaryMarkdown`
- [x] Live-verified against a throwaway Postgres + Redis instance:
  - the exact query/mapping the `export.md` route uses produces correct Markdown for a real seeded order (customer, address, note, item rows, totals all verified against the source data)
  - the `export-zip` route's dedup + `requestZipDownload()` call creates a correctly gallery-scoped `ZipDownload` record with the right file count, and reuses the cached record on a repeat request for the same order instead of building twice


markusthiel:
The Markdown export is a nice call — the CSV can't carry order-level context, and `isOrderSummaryAddress` validating the Json column instead of casting it is the right instinct.

Reusing the existing async ZIP pipeline rather than building a second one is also right, and it means this inherits the `force_zip64` fix from v0.73.5 for free. Deduping `fileIds` across lines is a detail that would have been easy to miss.

One thing I want to raise, and it isn't your bug — you've just landed on top of it.

The ZIP is created with `accessId: null`, and `requestZipDownload` defaults `variant` to `"original"`. The customer-facing route `GET /g/:slug/download/zip/:zipId` gates on

```ts
if (zip.accessId && zip.accessId !== visitor.accessId) → 403
```

so a null `accessId` passes for *any* visitor of that gallery. Net effect: a print-order ZIP of original files is reachable from the customer route, including in a gallery where original downloads are switched off for customers. It needs the `zipId`, which isn't guessable, so this is a door held shut by an unguessable id rather than by a check.

The existing gallery-wide Studio ZIP has exactly the same shape, so this predates your PR and I'd rather fix the pattern than block you on it. But this PR does widen it — a per-order bundle assembled to hand to a lab is a studio-internal artifact, and it now lives in a namespace the customer route reads.

Two ways I can see: mark studio-created ZIPs explicitly (a flag, or a sentinel `accessId`) and have the customer route reject those, or make the customer route require a non-null `accessId` match outright. The second is stricter and probably breaks the anonymous-public-gallery ZIP case, so I lean towards the first. Do you have a view? Happy to file it separately and keep it off this PR — say which you'd prefer.

manuzzi:
Good catch, and agreed it's a pre-existing pattern rather than something new here. I lean towards option 1 as well — a null `accessId` legitimately means "public/anonymous" for the customer's own whole-gallery download, so requiring a non-null match would break that case, and marking studio-created ZIPs explicitly is the distinction actually missing today.

Opened #52 against `main` (built on this branch since the print-order export-zip call site lives here, but as its own PR rather than commits onto this thread, per your note on #45 about not burying it): adds `ZipDownload.source` ("customer" | "studio"), rejects non-customer rows outright in the customer-facing route, and — since I was in there anyway — also re-checks `downloadEnabled`/`downloadOriginalsEnabled` against the gallery's live settings at read time rather than only at creation.

This PR's `export-zip` call site is tagged `source: "studio"` in #52, so once that lands the gap you flagged here is closed too.

PR #50, opened 2026-09-15
fix: allow a gallery expiry date to be cleared

bradley-varol:
# fix: allow a gallery expiry date to be cleared

`PATCH /galleries/:id` with `{"expiresAt": null}` failed, so an expiry date could not be removed once set.

The update mapping has always been written as `expiresAt: body.expiresAt ? new Date(body.expiresAt) : null`, so clearing was plainly the intended behaviour. The schema was simply inherited unchanged from `createGallerySchema`, where `expiresAt` is a non-nullable optional, and zod rejected the `null` before the mapping ever ran.

This makes it nullable on the **update** schema only. On create, `null` and omitted would mean the same thing, so that contract is untouched. Omitted still means "leave it alone", `null` now means "clear it" — the distinction the mapping already relied on.

Verified against a local stack: setting an expiry returns 200, clearing it now returns 200 and the column is `NULL` (it was a 500 before), and an invalid value is still rejected. Adds five schema tests, following the copy-the-schema convention in `sections.test.ts`.

### One thing I deliberately left alone

The rejection surfaced as a **500, not a 400**. `setErrorHandler` keys off `err.validation`, which is Fastify's own schema validation, so a `ZodError` thrown by `.parse()` has neither that nor a `statusCode` and falls through to 500. That affects every route using `schema.parse()`, not just this one, so it felt like its own change rather than something to widen this fix into. Happy to open a separate PR if you'd like it.


markusthiel:
Merged in v0.76.1, thanks.

The part I liked most is what you left out. You saw that the 500 comes from `setErrorHandler` only knowing Fastify's own validation, so a `ZodError` from `.parse()` falls straight through — and you recognised that as a separate change touching every route rather than widening this one. That's exactly the right call, and yes, please open that PR. It's a real bug: a client sending a malformed body currently gets told the server broke, when it should be told what it sent wrong.

One thing worth checking when you do: some routes may already rely on the 500 accidentally — a frontend `catch` that treats any non-2xx the same way won't care, but anything matching on the status code specifically would. `error-i18n.ts` on the frontend side is where I'd look first.

PR #51, opened 2026-09-15
fix: use singular wording for a one-file gallery count

bradley-varol:
A client gallery holding exactly one file rendered "1 FILES" in the header. The count printed a fixed plural noun, where the rest of the interface picks between an `Sg` and a `Pl` key at the call site.

Uses that existing convention rather than introducing a pluralisation mechanism.

Two translation problems surfaced while splitting the key. The German entry was still the English "Files", and the Finnish was "Tiedostot", a nominative plural that does not follow a numeral in Finnish — "2 tiedostoa" is the partitive. Both are corrected here since the key was being replaced anyway. Italian "file" is invariable, so both forms are deliberately identical.


markusthiel:
Merged in v0.76.1, thanks — and for the Finnish partitive in particular. That's the kind of thing a native speaker notices and nobody else does; the previous wording had been sitting there since the translation landed.

The German "Files" was mine, for the record. It came in with the first i18n pass and never got a second look because the string is uppercase and small enough to read as intentional.

PR #54, opened 2026-09-17
fix(api): return 400 for invalid request data, and make the error handler reachable

bradley-varol:
The follow-up you asked for in #50.

Sending a value the API rejects — an over-long title, an unknown status, a malformed date — answered **500** instead of 400, with the raw zod issue array in the response message. A caller could not tell its own bad request from a broken server.

## The status code was the symptom

While writing the `ZodError` branch I found the handler was not running at all. `setErrorHandler` sits **below** the route registrations in `server.ts`, and a Fastify error handler is only inherited by child contexts created *after* it is set. The `/api/v1` scope is registered above it, so nothing in the handler ever ran for an API request — including the `err.validation` branch that was already written. Errors were being formatted by Fastify's default handler the whole time, which is why the zod output ended up on the wire.

So the handler moves above the routes, and gains a `ZodError` branch. `parse()` throws a `ZodError`, which carries neither `err.validation` (Fastify's own schema validation, a different thing) nor a `statusCode`, so it would otherwise reach the catch-all. Issues are flattened to `path` / `code` / `message` rather than passed through, since a raw `ZodIssue` carries internals a caller has no use for and changes shape between zod versions.

## ⚠️ This changes the shape of error bodies

Worth your decision, not just a note. Because the handler never ran, errors came out in Fastify's default format, which includes a `statusCode` field in the body. They now use the format your handler already specified:

```
before  {"statusCode":500,"error":"Internal Server Error","message":"[{...raw zod...}]"}
after   {"error":"validation_failed","message":"request validation failed",
         "details":[{"path":"title","code":"too_big","message":"String must contain at most 200 character(s)"}]}
```

HTTP status codes are unchanged and were always the reliable signal. One knock-on: the `error` field now carries the exception name, so an unauthorised request reports `UnauthorizedError` where the default said `Unauthorized`. Say the word if you would rather I preserve the old strings.

## Verification

Checked against a running instance, not only in tests. The two requests that returned 500 now return 400 with the offending field named, valid requests are unaffected, and errors that are not validation failures reach the handler at all now.

Five tests, covering the 400 and its body shape, that a genuine fault still returns 500 without leaking its message, and — the one that matters here — that the handler applies to a route registered inside a prefixed scope. Anything exercising the handler directly would have passed for as long as it was unreachable.

Full API suite: 279 passed, 0 failed. `tsc --noEmit` and `npm run build` clean. Changelog entry added under Unreleased, including an upgrade note about the response shape.


markusthiel:
Merged in v0.77.1, thanks — and this was a bigger find than the title suggests. The status code was the symptom; the handler having never run is the actual bug, and it has a security side you didn't call out, so I added it to the changelog under its own heading: with the handler unreachable, genuine faults went out in Fastify's default body, which carries `err.message`. Prisma errors carry table and column names, filesystem errors carry paths. That's been going to clients since the handler was written.

Your question on the `error` field: no need to preserve the old strings. The frontend matches on its own snake_case codes (`not_found`, `selection_limit_reached`, …), which routes send directly and which never pass through this handler. Nothing on our side reads `Unauthorized` as a string, and the HTTP status was always the signal. I've said so in the upgrade note.

One small thing, not worth a follow-up on its own: the `ZodError` branch returns before `app.log.error`, the `err.validation` branch after it. So a Zod rejection is never logged, while a Fastify schema rejection is logged at error level. I'd put both at `warn` — a client sending bad input isn't an error on our side, but it's also not something I'd want to be completely blind to. If you touch that file again, fold it in; otherwise I'll do it when I'm next there.

The test that matters is the prefixed-scope one. Every test that exercised the handler directly would have kept passing for as long as it was unreachable — that's why nobody noticed.

